PR #16, opened 2019-09-20
Added basic localization support.

g-zachar:
Adds basic localization tool and exposes it to ux namespace.

### DIFF
--- a/js/src/locale.js
+++ b/js/src/locale.js
@@ -37,15 +37,20 @@
  */
 
 export class Locale {
+
+    constructor() {
+        this.__enabled = false;
+    }
+
     /**
      * Loads translation object from external json file.
      *  
      * @param {String} path Path to resource.
      * @return {Promise}
      */
-    static async load(path) {
+    async load(path) {
         if (!this.__enabled) {
-            Promise.resolve()
+            return;
         }
 
         await fetch(path)
@@ -60,7 +65,7 @@ export class Locale {
      *
      * @param {String} lang 
      */
-    static setLanguage(lang) {
+    setLanguage(lang) {
         this.__enabled = true;
         this.language = lang;
     }
@@ -70,7 +75,7 @@ export class Locale {
      *
      * @return {Object}
      */
-    static get tr() {
+    get tr() {
         return this.__trObj[this.language];
     }
 
@@ -79,7 +84,7 @@ export class Locale {
      *
      * @param {Object} trObj 
      */
-    static loadFromObject(trObj) {
+    loadFromObject(trObj) {
         this.__trObj = trObj;
         for (const lang of Object.values(this.__trObj)) {
             for (const str of Object.keys(lang)) {
@@ -88,8 +93,6 @@ export class Locale {
         }
     }
 }
-
-Locale.__enabled = false;
 
 /**
  * Extended string class used for localization.
@@ -106,10 +109,7 @@ class LocalizedString extends String {
      * @param  {...any} args List of arguments for placeholders.
      */
     format(...args) {
-        let sub = this;
-        args.forEach((arg, index) => {
-            sub = sub.split(`{${index}}`).join(arg);
-        });
+        const sub = args.reduce((string, arg, index) => string.split(`{${index}}`).join(arg), this);
         return new LocalizedString(sub);
     }
 }
--- a/js/src/locale.js
+++ b/js/src/locale.js
@@ -44,6 +44,10 @@ export class Locale {
      * @return {Promise}
      */
     static async load(path) {
+        if (!this.__enabled) {
+            Promise.resolve()
+        }
+
         await fetch(path)
         .then((resp) => resp.json())
         .then((resp) => {
@@ -57,6 +61,7 @@ export class Locale {
      * @param {String} lang 
      */
     static setLanguage(lang) {
+        this.__enabled = true;
         this.language = lang;
     }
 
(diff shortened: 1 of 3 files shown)


js/src/locale.js
```diff
@@ -0,0 +1,108 @@
...
+export class Locale {
+    /**
+     * Loads translation object from external json file.
+     *  
+     * @param {String} path Path to resource.
+     * @return {Promise}
+     */
+    static async load(path) {
+        await fetch(path)
+        .then((resp) => resp.json())
```
erikhaandrikman: We can remove the then chain by: 

```
const response = await fetch(path)
const obj = await response.json();
this.loadFromObject(obj);
```

js/src/locale.js
```diff
@@ -0,0 +1,108 @@
...
+     * Replaces each placeholder value (e.g. {0}, {1}) with corresponding argument.
+     * 
+     * E.g.:
+     * > new LocalizedString('{0} and {1} and {0}').format('A', 'B');
+     * A and B and A
+     *
+     * @param  {...any} args List of arguments for placeholders.
+     */
+    format(...args) {
+        let sub = this
```
erikhaandrikman: missing semicolon

erikhaandrikman:
Do we want to output a warning when someone tries to set a non-existing language?

js/src/locale.js
g-zachar: That's true but this creates local variables unnecessarily. Another alternative would be await expressions, but they cannot be chained nicely and are even less readable.

erikhaandrikman: fair point, lets just leave it as is

js/src/locale.js
g-zachar: Fixed

js/src/locale.js
```diff
@@ -0,0 +1,115 @@
...
+export class Locale {
+    /**
+     * Loads translation object from external json file.
+     *  
+     * @param {String} path Path to resource.
+     * @return {Promise}
+     */
+    static async load(path) {
+        if (!this.__enabled) {
+            Promise.resolve()
```
rbielawski: Shouldn't it be `return Promise.resolve();` to prevent execution of next lines?
Or even just `return;`, since it's an async function that doesn't need to return promises explicitly.

js/src/locale.js
```diff
@@ -0,0 +1,115 @@
...
+     * Replaces each placeholder value (e.g. {0}, {1}) with corresponding argument.
+     * 
+     * E.g.:
+     * > new LocalizedString('{0} and {1} and {0}').format('A', 'B');
+     * A and B and A
+     *
+     * @param  {...any} args List of arguments for placeholders.
+     */
+    format(...args) {
+        let sub = this;
```
rbielawski: I believe this reference is redundant if you're using arrow function in next line.

js/src/ux.js
```diff
@@ -4,11 +4,13 @@ import Ui from "./Ui.js";
...
 
 import tools from "./tools/tools.js";
+import { Locale } from "./locale.js";
 
 const ux = {
     Ui,
     App,
-    tools
+    tools,
+    locale: Locale,
```
rbielawski: What is the purpose of changing name to lowercase here?

js/src/locale.js
g-zachar: Yep, I'll correct this.

js/src/locale.js
g-zachar: Can you elaborate on this further?

js/src/ux.js
g-zachar: Following camelCase variable naming convention.

js/src/locale.js
rbielawski: I mean you could just use `this` inside arrow function, it will not create a new context:
`this = this.split(`{${index}}`).join(arg);`

js/src/ux.js
rbielawski: Then why are you exporting/importing it as `Locale`? It's inconsistent.

rbielawski: If it's a class, then i believe it should be named `Locale`, if it's not - it should be `locale`, in all places.

js/src/locale.js
g-zachar: `this` reference is immutable, trying to assign value for it gives `Uncaught SyntaxError: Invalid left-hand side in assignment`

js/src/ux.js
g-zachar: I disagree. Keep in mind that ux.locale is a variable which can have anything assigned to it. Example scenario would be having 2 different localization libraries e.g. `Locale` and `DifferentLocale` that implement same API. Now, you could seemlessly switch between the two, because the `ux` API wouldn't change, but if you named variable same as class import, then you would break compatibility by making the `ux.DifferentLocale` switch.

rbielawski: I see. Ok, let's just leave it as it is now.

js/src/locale.js
rbielawski: Oh, you're absolutely right, my mistake, sorry :).

However still, I'd try to avoid creating and mutating this kind of 'helper' variables and rather make use of map/reduce etc. In this case it could be something like:
```
format(...args) {
    const sub = args.reduce((string, arg, index) => string.split(`{${index}}`).join(arg), this);
    
    return new LocalizedString(sub);
}
```
I think it's more clear, but I can agree it's just a preference, so I do not expect you to do it this way. Up to you.

PR #58, opened 2020-03-19
App: Metric and closure changes

wouterlucas:
Split out closeApp and the cleanup, this allows the bootstrapper to call
close externally. 

Add this.destroy() to trigger lightning cleanup, so we do some GC before
exiting. 

Move launch/close metric to bootstrapper.

### DIFF
--- a/src/Application/index.js
+++ b/src/Application/index.js
@@ -33,7 +33,6 @@ if (window.innerHeight === 720) {
 }
 
 export default function(App, appData, platformSettings) {
-  Metrics.app.launch()
   return class Application extends Lightning.Application {
     constructor(options) {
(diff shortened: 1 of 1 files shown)


src/Application/index.js
```diff
@@ -33,7 +33,6 @@ if (window.innerHeight === 720) {
 }
 
 export default function(App, appData, platformSettings) {
-  Metrics.app.launch()
```
michielvandergeest: @wouterlucas we are sure that we don't want to track the `app.launch` metric anymore? just confirming since we still do have `app.close` in the close method

wouterlucas: Oh! Close should be gone also, oversight on my end. Thanks for catching that.

PR #59, opened 2020-03-19
Feature/profile defaults

jvanleest82:


src/Profile/helpers.js
```diff
@@ -0,0 +1,59 @@
...
+    const geoLocationError = error => {
+      return resolve(defaultValue)
+    }
+    const geoLocationOptions = {
+      enableHighAccuracy: true,
+      timeout: 5000,
+      maximumAge: 0,
+    }
+
+    if ('geolocation' in navigator) {
```
michielvandergeest: If for some reason geolocation doesn't exist, the `getLanLon`-method will not return anything. So you probably want to resolve the defaultValue in this case.

src/Profile/index.js
```diff
@@ -1,5 +1,14 @@
-let getInfo = () => Promise.resolve()
-let setInfo = () => Promise.resolve()
+import Settings from '../Settings/index'
+import { defaultProfile } from './defaults'
+
+let getInfo = key => {
+  const profile = { ...defaultProfile, ...Settings.get('app', 'profile') }
```
michielvandergeest: I think profile should be a "platform" setting, instead of an "app" setting

src/Profile/defaults.js
```diff
@@ -0,0 +1,18 @@
...
+  language: getLanguage('en'),
+  latlon: getLatLon([40.7128, 74.006]),
+  locale: getLocale('en-US'),
+  mac: '00:00:00:00:00:00',
+  operator: 'Metrological',
+  platform: 'Metrological',
+  packages: [],
+  uid: 'ee6723b8-7ab3-462c-8d93-dbf61227998e',
+  stbType: 'Metrological',
+}
```
michielvandergeest: @wouterlucas @erikhaandrikman @woutermeek any opinion about these default values? Are they the best "sensible defaults"?

package-lock.json
```diff
@@ -1,5524 +0,0 @@
-{
```
michielvandergeest: package-lock.json shouldn't have been deleted :)

src/Profile/defaults.js
wouterlucas: I'd prefer San Francisco over New York (just kidding). No they look good.

Generic question though, will the meaning of these values be documented somewhere? Will we rely on a fixed set of values?

To elaborate, the "ageRating" triggered me. I earlier had an `integer` for that, or number, to define age. Now the default is set to `adult`, what other values can a developer expect? We'll need to document what these values mean and what can a developer expect for each one of them.

Granted, we may pull that apart of this PR and do the doc update separately. To avoid making the docs a long pole and block this PR.

src/Profile/helpers.js
jvanleest82: Added a fallback to defaultValue

src/Profile/index.js
jvanleest82: Changed "app" to "platform"

src/Profile/defaults.js
michielvandergeest: I actually think we should make a habit out of adding documentation as part of each feature PR (if there are changes that need documentation).

In this case we should probably document the default values. But also how they can be customized (during development) via settings.json.

jvanleest82: I have updated the documentation with the default values added.

PR #62, opened 2020-04-14
Feature/locale

jvanleest82:


src/Application/index.js
```diff
@@ -52,7 +52,7 @@ export default function(App, appData, platformSettings) {
     _setup() {
       Promise.all([
         this.loadFonts((App.config && App.config.fonts) || (App.getFonts && App.getFonts()) || []),
-        Locale.load((App.config && App.config.locale) || (App.getLocale && App.getLocale())),
+        initLocale(),
```
michielvandergeest, line 55: This means that the Locale plugin is _always_ initialized (and tries to read locale files - that might not exist), even if the App doesn't implement  Localization

src/Locale/index.js
```diff
@@ -1,116 +1,38 @@
...
+export default {
+  get(attributeName, ...args) {
+    if (!(loadedLanguageFile && loadedLanguageFile[attributeName])) return attributeName
+    if (args && args.length > 0) return format(loadedLanguageFile[attributeName], args)
+    return loadedLanguageFile[attributeName]
+  },
+
+  // Set ISO locale code aka: 'en-US, nl-NL, etc' default is en-US.
+  setLocale(locale) {
+    return loadTranslationFile(isCorrectLocale(locale, defaultLocale), defaultLocale).then(
```
michielvandergeest, line 27: I would move the call to `isCorrectLocale` to the `loadTranslationFile` method itself (then you can also remove the import of isCorrectLocale)

src/Locale/index.js
```diff
@@ -1,116 +1,38 @@
...
+      languageFile => {
+        if (languageFile) {
+          loadedLanguageFile = languageFile
+          return true
+        }
+        console.warn('Locale: could not set locale: ' + locale)
```
michielvandergeest, line 33: Better to use the `Log` class so logs can be turned off in production.

src/Locale/index.js
```diff
@@ -1,116 +1,38 @@
...
+
+const defaultLocale = 'en-US'
+let loadedLanguageFile = undefined
+
+// Init
+export const initLocale = async () => {
+  return Profile.locale().then(locale => {
+    return Locale.setLocale(locale).then(isSet => {
+      return isSet
+    })
```
michielvandergeest, line 13: Consider adding a catch

src/Locale/index.js
```diff
@@ -1,116 +1,38 @@
...
-   */
-  loadFromObject(trObj) {
-    this.__trObj = trObj
-    for (const lang of Object.values(this.__trObj)) {
-      for (const str of Object.keys(lang)) {
-        lang[str] = new LocalizedString(lang[str])
-      }
-    }
-  }
+import Locale from './'
```
michielvandergeest, line 1: This is a bit weird, to import yourself ;)

src/Locale/index.js
```diff
@@ -1,116 +1,38 @@
...
+import { loadTranslationFile, format, isCorrectLocale } from './helpers'
+
+const defaultLocale = 'en-US'
+let loadedLanguageFile = undefined
+
+// Init
+export const initLocale = async () => {
+  return Profile.locale().then(locale => {
+    return Locale.setLocale(locale).then(isSet => {
+      return isSet
```
michielvandergeest, line 12: how / where do you use this return value?

src/Locale/helpers.js
```diff
@@ -0,0 +1,45 @@
...
+    const integerWithinBrackets = new RegExp(`\\{\\s*(${index})\\s*}`)
+    return previousText.replace(integerWithinBrackets, currentText)
+  }, text)
+}
+
+// Returns locale if naming is correct, else returns default locale
+export const isCorrectLocale = (locale, defaultLocale) => {
+  const isLocale = new RegExp('^([a-zA-Z][a-zA-Z]-[a-zA-Z][a-zA-Z])(?!.)')
+  if (isLocale.test(locale)) return formatLocale(locale)
+  console.warn(`Locale: ${locale} is invalid, switching to default`)
```
michielvandergeest, line 36: Better to use Log plugin, so logs can be turned off in production.

src/Locale/helpers.js
```diff
@@ -0,0 +1,45 @@
+import Utils from '../Utils/index'
+
+export const loadTranslationFile = (isoLocale, defaultLocale) => {
+  const localePath = Utils.asset(`locale/${isoLocale}.json`)
+  const localeDefaultPath = Utils.asset(`locale/${defaultLocale}.json`)
+
+  return fetch(localePath).then(response => {
+    if (response.ok) return response.json()
+    console.warn(`Locale: ${isoLocale}.json - ${response.statusText}`)
+    console.warn(`Locale: Defaulting to /${defaultLocale}.json`)
```
michielvandergeest, line 10: Better to use Log plugin so it can be turned off easily in production

src/Application/index.js
jvanleest82: `static get locale(){
    return true
}`

Needs to be added to App.js to init locale

src/Locale/index.js
jvanleest82: Moved to `loadTranslationFile`

src/Locale/index.js
jvanleest82: Changed to `Log` class

src/Locale/index.js
jvanleest82: Removed `.then` and added catch to the `fetch` of `loadTranslationFile`

src/Locale/index.js
jvanleest82: Weird indeed :) , removed.

src/Locale/index.js
jvanleest82: Useless return value, removed.

src/Locale/helpers.js
jvanleest82: Changed to `Log` class

src/Locale/helpers.js
jvanleest82: Changed to `Log` class

src/Application/index.js
jvanleest82: Update:

`static get locale(){ return true } ` Will load browser default language.
`static get locale(){ return false }` Will not load `Locale` on init.
`static get locale(){ return 'en-CA' }` Will load preferred language on init ('en-CA' for example)

michielvandergeest:
hey @jvanleest82, since the move to the rdkcentral repository it's required for all contributors to sign the License Agreement. The PR is blocked for merging, could you verify if you've signed it? Thanks

chiefcll:
Looking to get this going today. Any progress would be appreciated 👍

chiefcll:
🦗

michielvandergeest:
Sorry. I missed your previous comment. This PR still needs some development cycles before it will make it into master. Hope to free up some time for that in the next 2 weeks. In its current state it's an improvement on the current Locale plugin yes, but IMO not mature enough to be "the new" Locale implementation.

chiefcll:
The current Locale plugin is not usable. I've spent a few hours wiring it in due to how it interacts with Launch code. Trying to setLanguage before calling Launch without a locale config makes it fetch undefined. After figuring that out and getting the JSON to load, if you try to patch with `Locale.tr.HOME_TITLE` - that is not a string, its a `LocalizedString` which JS treats as an Object and Lightning gives a `Settings must be object literal`. So now I have to do `title: String(Locale.tr.HOME_TITLE),`

michielvandergeest:
Agreed. The Locale is a leftover from the previous SDK version that comes with (serious) flaws. So when we release the successor of the Locale we should get it right. And the current version isn't there yet.

Good to hear you guys are also considering to use the Locale from the SDK, that helps decide how we allocate our dev time. 

I believe the following snippet could help to get the current Locale running for now:

```js
// src/App.js
export default class App extends Lightning.Component {

    static getLocale() {
        Locale.__enabled = true // hack
        Profile.language().then(language => {
            Locale.setLanguage(language || 'en');
        })
        return Utils.asset('locale/locale.json');
    }
}
```

PR #64, opened 2020-04-18
Added a realtime FPS counter (inspired by @erikhaandrikman and @woute…

michielvandergeest:
…rlucas) that can be toggled via platformSetting 'showFps'.

### DIFF
--- a/src/FpsCounter/index.js
+++ b/src/FpsCounter/index.js
@@ -1,0 +1,66 @@
+import Lightning from '../Lightning'
+
+export default class FpsIndicator extends Lightning.Component {
+  static _template() {
+    return {
+      rect: true,
+      color: 0xffffffff,
+      texture: Lightning.Tools.getRoundRect(80, 80, 40),
+      h: 80,
+      w: 80,
+      x: 100,
+      y: 100,
+      mount: 1,
+      Background: {
+        x: 3,
+        y: 3,
+        texture: Lightning.Tools.getRoundRect(72, 72, 36),
+        color: 0xff008000,
+      },
+      Counter: {
+        w: w => w,
+        h: h => h,
+        y: 10,
+        text: {
+          fontSize: 32,
+          textAlign: 'center',
+        },
+      },
+      Text: {
+        w: w => w,
+        h: h => h,
+        y: 48,
+        text: {
+          fontSize: 15,
+          textAlign: 'center',
+          text: 'FPS',
+        },
+      },
+    }
+  }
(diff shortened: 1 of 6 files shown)


src/FpsCounter/index.js
```diff
@@ -0,0 +1,66 @@
...
+  _setup() {
+    this.fps = 0
+    this.stage.on('frameStart', () => {
+      this.fps = ~~(1 / this.stage.dt)
+    })
+    this.tag('Counter').loadTexture()
+    this.interval = setInterval(this.showFps.bind(this), 100)
+  }
+
+  _detach() {
```
erikhaandrikman, line 51: I think we should also cleanup the `frameStart` listener on `detach`

src/FpsCounter/index.js
```diff
@@ -0,0 +1,66 @@
...
+    }
+  }
+
+  _setup() {
+    this.fps = 0
+    this.stage.on('frameStart', () => {
+      this.fps = ~~(1 / this.stage.dt)
+    })
+    this.tag('Counter').loadTexture()
+    this.interval = setInterval(this.showFps.bind(this), 100)
```
wouterlucas, line 48: I'd raise the 100ms update a little bit. To avoid getting lower FPS for updating the FPS every 100ms :)
I assume 300 - 500ms is probably enough to get a decent feeling of how well the FPS is updating when rendering frame.

src/FpsCounter/index.js
```diff
@@ -0,0 +1,66 @@
...
+    clearInterval(this.interval)
+  }
+
+  showFps() {
+    // green
+    let bgColor = 0xff008000
+    // orange
+    if (this.fps <= 40 && this.fps > 20) bgColor = 0xffffa500
+    // red
+    else if (this.fps <= 20) bgColor = 0xffff0000
```
wouterlucas, line 61: nice :)

src/FpsCounter/index.js
michielvandergeest: Good point. Added in `this.stage.off()` .. that's should be sufficient?

src/FpsCounter/index.js
michielvandergeest: Raised the default to 300ms, but added configuration option that enables to set any custom ms.

src/FpsCounter/index.js
erikhaandrikman: yeah should be fine

PR #66, opened 2020-04-21
Feature/purchase

Thomvl:
Todo: Get correct app name


wouterlucas (commented):
I think we need to clarify this a little, initial remarks form my side:

- payment object or the params should be more well defined. Ideally the SDK does as much as we can to keep the complexity of the billing implementation to a mimimum for the dev. For example we should be able to generate a payment object or template, the SDK should allow of way to generate an example so its not too complex. And we should specify what the contents of the payment object is.
- requires documentation of what purchases are and how they work. This needs to be added to the PR
- we want to support app, in-app and subscription type of purchases

michielvandergeest:
Closing in favour of `feature/purchase` branch

PR #70, opened 2020-04-28
Feature: Add setter and subscribe to Settings

chiefcll:
For our application we have platform settings that we get async - we'd like to run them through the Settings from SDK. We also need to listen for those changes for our app to update.


src/Settings/index.js
```diff
@@ -13,4 +18,15 @@ export default {
   has(type, key) {
     return settings[type] && key in settings[type]
   },
+  set(key, value, type = 'user') {
+    settings[type] = settings[type] || {}
+    settings[type][key] = value
+    publish(type, key, value)
+  },
+  subscribe(callback) {
```
michielvandergeest, line 26: I think it would be nicer if we could pass a `key` to the `subscribe` method, so the callback only gets invoked whenever that specific setting changes.

chiefcll: Pondering it - Redux follows the subscribe to store changes and you figure out what changed philosophy. Maybe we can flatten the Settings to be only two levels deep and drop the "type", otherwise this becomes `subscribe(type, key, callback)` which is strange since app / platform can't change.

michielvandergeest: I think we should omit the possibility to pass a type all together. Although I'd like to keep storing user defined settings under the user-type. So that basically becomes a hardcoded type.

And since only user defined settings ever call the publish method, I don't think there's a need to explicitly pass a type here either.

My suggestion would actually be to store the callbacks with the key they subscribe to, so we can prevent from looping through all registered callback when any key changes.

chiefcll: How would you change the getter?

```
get(type, key) {
    return settings[type] && settings[type][key]
  },
  has(type, key) {
    return settings[type] && key in settings[type]
  },
```

Would someone have to do `get('user', 'key')` - how would they know the type user? would that be part of the docs?

michielvandergeest: I would keep the `get`-method _as is_ and have developers pass 'user' as a `type`. That would be necessary anyway, even in the case where you'd set a setting without explicitly passing a type.

And yes, agreed this would have to be properly documented. But it does make sense that you'd 'get' user-defined settings by passing 'user' a type.

chiefcll:
All done - LMK

src/Settings/index.js
```diff
@@ -1,13 +1,14 @@
 const settings = {}
-let subscribers = []
+let subscribers = {}
```
michielvandergeest, line 2: We can change this to a `const` now, because we're not reassigning anymore in `clearSubscribers()`.

src/Settings/index.js
```diff
@@ -18,19 +19,16 @@ export default {
...
   },
-  subscribe(callback) {
-    subscribers.push(callback)
+  subscribe(key, callback) {
+    subscribers[key] = subscribers[key] || []
+    subscribers[key].push(callback)
   },
-  clearSubscribers() {
-    subscribers = []
+  clearSubscribers(key) {
```
michielvandergeest, line 31: I think we still need to document this functionality

chiefcll:
Nice catch - all done

michielvandergeest:
hey @chiefcll, since the move to the rdkcentral repository it's required for all contributors to sign the License Agreement. The PR is blocked for merging, could you verify if you've signed it? Thanks

chiefcll:
Thanks for letting me know - didn't see the email - signed now.

michielvandergeest:
@chiefcll I've added some stuff and refactored some things. Appreciated if you could take it for a spin and see if it all works well for you so we can get on with merge the PR

PR #93, opened 2020-07-03
Feature/image quality setting

michielvandergeest:
- new image quality setting
- this is a nested setting (i.e. quality.imageSize) so there is now support for nested settings via 'dot'-notation

### DIFF
--- a/docs/plugins/image.md
+++ b/docs/plugins/image.md
@@ -85,7 +85,7 @@ You can however experiment with this during local development via `settings.json
 }
 ```
 
-The platform setting `quality.imageSize` can be a value between `1` and `100` (where 100 equals 100%, high quality)
+The platform setting `quality.imageSize` can be a value between `1` and `100` (where 100 equals 100% - high quality)
 
 ## Deprecated methods
 
--- a/docs/plugins/image.md
+++ b/docs/plugins/image.md
@@ -64,6 +64,29 @@ Generate an image without resizing it (i.e. use the original dimensions), while 
 Img(url).original()
 ```
 
+## Image quality
+
+To increase performance on lower end boxes, especially those with limited GPU memory, there exists a `platform` setting to control the _image quality_.
+Depending on this setting the images returned by the image server will be _smaller_ than actually displayed on the screen.
+Lightning will then _stretch_ the images to fit the desired dimensions. This will save on used GPU memory, but off course impacts the visual quality of the images.
+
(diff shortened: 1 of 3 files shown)


docs/plugins/image.md
```diff
@@ -64,6 +64,29 @@ Generate an image without resizing it (i.e. use the original dimensions), while
...
+Please note that within the context of the Metrological AppStore this setting is defined as a platform setting _outside_ of the control of the App.
+You can however experiment with this during local development via `settings.json`:
+
+```json
+{
+    "appSettings": {
+    },
+    "platformSettings": {
+        "quality": {
+          "imageSize": 50
```
wouterlucas, line 82: Nitpicking, but quality and having a suboption image makes sense, however the notion of "imageSize" with the word size throws me off as that might imply something else with the value you need to provide (as in width / hight) as input.

I'd vouch for, either a specific imageQuality tag since we dont have any other quality modifiers at this point. 
Or `quality : { image : 50 }` without the Size word in it to avoid confusion.

In the future we might use the same image Quality indicator to toggle more, such as encoding / compression and the lot. I would imagine that is all based off of the same "image quality" indicator? Or would we do a separate "switch" for all of those and bubble the extra complexion up to the platformSettings?

michielvandergeest: +1 for thinking well about naming :)

Let’s separate in 2 topics: the naming of “size” and the grouping of settings:

`imageSize` was chosen because in fact it causes te Image Server to return a smaller _size_ image (by the percentage specified), which in turn off course has an effect on the quality (due to the stretching).

As for grouping: I think we really need to start grouping settings, otherwise things will become too messy.

The initial idea was to have `quality` group and add more related options such as `animations` and `shaders` to it. Not sure if quality is the best key name though ...

But your comment about encoding / compression, made me think that maybe we need to take it one level deeper even, i.e.:

```
{
	quality: {
		image: {
			size: ’50%’,
			compression: true,
		},
		animations: {
		}
	}
}
```

I like the idea of having the option to control everything with it’s own setting, but I also see value in a general quality setting (i.e. low, medium, high) that then presets all these fine grained options (with the possibility to override).

@erikhaandrikman: your thoughts?

g-zachar: I suggest `scale` instead of `imageSize` and range from 0 to 1.

michielvandergeest: I kind of like `scale` as well, but after discussion with @wouterlucas and @jansenroel we've decided to go with `image.quality` and keep the range of 1-100, matching how you can set quality on jpeg images for example.

PR #94, opened 2020-07-08
Feature/Announcer

chiefcll:
Add A11y announcer class with browser based text to speech

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules
 .idea
 support/lib
 support/polyfills
+dist/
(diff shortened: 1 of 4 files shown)


wouterlucas:
Nice! Thanks @chiefcll, much appreciated.

@michielvandergeest I think the only thing we need to consider is the ability to overwrite the SpeechSynthesis engine from the bootstrapper. Ideally we should rely on the one configured in the browser, but that might not work for all cases. Of course we could polyfill the window.SpeechSynthesis or we could add in a platform setting that sets the desired speech engine. Wdyt?

michielvandergeest:
hey @chiefcll thanks for your contribution :)

I did a quick first review yesterday together with @erikhaandrikman. And one of our concerns indeed was the (hardcoded) dependency on `speechSynthesis`. Ideally we would make that more configurable and / or injectable as @wouterlucas mentioned.

For further review we need to take some time play around with the implementation and test it in practice. By any chance do you have something like a test app that implements the full Announcer functionality that you are able to share? That would be helpful!

Thanks again!

chiefcll:
Play around with it LMK. FYI - Announcer class is created with a Speech function passed in, so it is overridable. I just added a default for Web devs.

erikhaandrikman:
@chiefcll i think we can close this PR since it's already part of the ui-components repo, right?

chiefcll:
Announcer should be part of the SDK - its more a util like locale. It should be added once to an application to provide a11y and should be a standard feature for all Lightning apps.

PR #97, opened 2020-07-14
Dev

michielvandergeest:


src/Router/index.js
```diff
@@ -808,7 +815,6 @@ const getRouteByHash = hash => {
 
         // split regex and modifiers so we can use both
         // to create a new RegExp
-        // eslint-disable-next-line
```
michielvandergeest: this comment is necessary, otherwise the lint commit hook fails ...

PR #108, opened 2020-08-10
Feature/palette

jfboeve:
### Colors Library for calculating and storing colors so you can easily access them in your project.

The library comes with 8 different colors; white, black, red, green, blue, yellow, cyan, and magenta.

Getting a color from the Palette is as easy as getting an asset with the Utils api. For example:

`Colors.get('white')`

If your want your white color to have some opacity / alpha your can use the following: 

`Colors.get('white', 0.4)`

The other options to modify your color are, lighten and darken. For example:

`Colors.get('white', {darken: 0.3})`

If you want to darken AND alpha you can use:

`Colors.get('white', {darken: 0.3, alpha: 0.3})`

### Customizing

If you wish to store and use your own colors you can use the following function:

`Colors.add({
    focus: '#444444',
    background: '#222222'
});`

### Mixing

You can also mix to different colors:

`Colors.mix('white', 'black', 0.5)`


src/Colors/index.js
```diff
@@ -0,0 +1,101 @@
...
+}
+
+const calculateColor = (value, options) => {
+  let targetColor = normalizeColor(value)
+  if (!isNaN(options)) {
+    options = { alpha: options }
+  }
+  if (options && isObject(options)) {
+    if (!isNaN(options.darken)) {
+      targetColor = mergeColors(targetColor, 0xff000000, 1 - options.darken)
```
michielvandergeest, line 25: maybe we should check if options.darken is a number between 0 and 1?

src/Colors/index.js
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    if (!isNaN(options.lighten)) {
+      targetColor = mergeColors(targetColor, 0xffffffff, 1 - options.lighten)
```
michielvandergeest, line 29: same here, maybe check if it's a number between 0 and 1?

src/Colors/index.js
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    if (!isNaN(options.alpha)) {
+      targetColor = calculateAlpha(targetColor, options.alpha)
+    }
+  }
+  return targetColor || 0
+}
+
+const normalizeColor = color => {
```
michielvandergeest, line 43: I would suggest to rename this function to be a bit more descriptive of what it does. It wasn't instantly clear to me that this function converts hex colors to the argb format

src/Colors/index.js
```diff
@@ -0,0 +1,101 @@
...
+  let targetColor = colors[color]
+  if (!targetColor) {
+    targetColor = color
+  }
+  if (isString(targetColor) && /^#[0-9A-F]{6}$/i.test(targetColor.toUpperCase())) {
```
michielvandergeest, line 48: - I don't think we need `toUppperCase` since the regexp is already case insenstive
- maybe we should add support for `#333` as well (since it's so common in CSS, I guess it's very possible people will use the hex shorthand :) )

src/Colors/index.js
```diff
@@ -0,0 +1,101 @@
...
+    targetColor = `0xff${targetColor.slice(1)}` * 1
```
michielvandergeest, line 49: if we add grouping in the regexp, we can use that value right away to construct the color instead of slicing

src/Colors/index.js
```diff
@@ -0,0 +1,101 @@
...
+  }
+  return targetColor || 0xffffffff
+}
+
+const cleanUpNormalizedColor = color => {
+  for (let c in normalizedColors) {
+    if (c.indexOf(color) > -1) {
+      normalizedColors[c] = undefined
```
michielvandergeest, line 57: is it needed to set `undefined` since we delete the prop straight after?

src/Colors/index.js
```diff
@@ -0,0 +1,101 @@
...
+    }
+  }
+}
+
+export default {
+  get(value, options) {
+    // create color tag for storage
+    let tag = `${value}${options !== undefined ? JSON.stringify(options) : ''}`
+    // check if value is not a number, or if options is not undefined
+    if (isNaN(value) || options) {
```
michielvandergeest, line 68: do we need this check before we can return the stored color?

michielvandergeest (commented):
Looks good! See some comments inline

src/Colors/index.js
jfboeve: made a function for limiting the number within the range of 0 - 1

src/Colors/index.js
jfboeve: made a function for limiting the number within the range of 0 - 1

src/Colors/index.js
jfboeve: changed function name to "normalizeHexToARGB"

src/Colors/index.js
jfboeve: Reworked this part of the function, which removed the toUppercase and added support for 3 digit and converting it to 6 digit.

src/Colors/index.js
jfboeve: Did some changes, regarding the regex. Might need some changes still but i think we are close.

src/Colors/index.js
jfboeve: No we do not

src/Colors/index.js
jfboeve: removed it

src/Colors/index.js
michielvandergeest: @erikhaandrikman mind taking a look as our regexpert 😀

src/Application/index.js
```diff
@@ -183,6 +185,14 @@ export default function(App, appData, platformSettings) {
       return initLanguage(file, language)
     }
 
+    loadColors(config) {
+      let colors = Utils.asset('colors.json')
```
michielvandergeest, line 189: Nit-picky: but can we call the variable `file` (instead of `colors`), so it's in line with the loadLanguage function?

src/Colors/index.js
```diff
@@ -0,0 +1,129 @@
...
+      addColors(file)
+      resolve()
+    }
+    fetch(file)
+      .then(response => response.json())
+      .then(json => {
+        addColors(json)
+        resolve()
+      })
+      .catch(e => {
```
michielvandergeest, line 49: Maybe add a `Log.error()` here?

michielvandergeest:
Reopening since the PR was closed by accident when deleting the dev branch 😀

docs/plugins/colors/index.md
```diff
@@ -0,0 +1,79 @@
...
+	static colors() {
+        //default
+        return true
+        //object
+        return {
+            background: '#c9deff',
+            focus: '#276ee5'
+        }
+        //string
+        return 'my/custom/language-file.json'
```
michielvandergeest, line 32: my custom colors file :)

docs/plugins/colors/index.md
```diff
@@ -0,0 +1,79 @@
...
+    }
+}
+```
+
+### Colors file
+
+With the default, and string option during boot the Colors plugin expects a JSON file. The default option will look for a JSON file named colors.json. Or a JSON file at a custom location defined with the string option. The JSON file should look something like this:
+
+```js
```
michielvandergeest, line 41: json type

docs/plugins/colors/index.md
```diff
@@ -0,0 +1,79 @@
...
+{
+    "background": "#c9deff",
+    "focus": "#276EE5",
```
michielvandergeest, line 44: json can't have a dangling comma

docs/plugins/colors/index.md
```diff
@@ -0,0 +1,79 @@
...
+```
+
+You can also make your colors lighter or darker. For example:
+
+```js
+Colors('blue').lighter(0.3).get()
+Colors('green').darker(0.3).get()
+```
+
+With this library you can also chain methods, if for example a specific color darker and give it a specific opacity:
```
michielvandergeest, line 69: We use the term 'plugin' vs 'library'

docs/plugins/colors/index.md
```diff
@@ -0,0 +1,79 @@
...
+
+```js
+Colors('red').darker(0.2).alpha(0.8).get()
+```
+
+If you want to mix two different colors you can do it with the mix function. The first two parameters can be stored colors, #RRGGBB, or ARGB. The third parameter is the percentage of how it is mixed, we normalized it to a value from 0 to 1.
+
+```js
+Colors('cyan').mix(Colors('magenta').get(), 0.5).get()
+```
```
michielvandergeest, line 79: I'm still missing some documentation on the saturation methods? Other than that, looks great!

docs/plugins/colors/index.md
jfboeve: oops

docs/plugins/colors/index.md
jfboeve: Yeah, thought I had every library changed to plugin, also found it somewhere else and changed it accordingly.

docs/plugins/colors/index.md
jfboeve: I have rearranged some bits around in the docs to and added the Hue, Saturation, Lightness functionalities to the docs.

michielvandergeest (approved):

PR #111, opened 2020-08-12
Added TV plugin.

michielvandergeest:


docs/plugins/tv.md
```diff
@@ -0,0 +1,140 @@
+# TV
+
+The TV plugin serves as an abstraction layer to the _live TV_ functionality of a STB. The interface gives you access to the current channel
+and program information, and allows you to change the TV channel from inside an App.
```
wouterlucas, line 4: Might need to nuance this or elaborate. Changing TV channels from within an app is a restricted operation. Not every, or very likely most of the applications, will not be able to change the current Live TV Channel. However if the application is whitelisted to use this the functionality is available.

docs/plugins/tv.md
```diff
@@ -0,0 +1,140 @@
...
+The `entitled`-method returns a _promise_ which returns `true` when the user is entitled and `false` when not.
+
+During _development_ the `entitled`-method returns the entitlement value of the random mocked channel. Optionally you can [overwrite](#overwriting-default-values) the default values.
+
+### ChangeChannel
+
+Changes the current channel on the TV.
+
+```js
+TV.changeChannel(number).then(newChannel => console.log(newChannel))
```
wouterlucas, line 78: why not make this the same as the .channel API but the only difference is we're providing a number to actually manipulate the active channel? Or would that simplify things too much from a app perspective OR complicate things on the internal side?

michielvandergeest: I've actually considered that yes and it's very simple to make that work on the internal side.
I think it looks nice, so shall we change the API to work like that (getter setter style)?

docs/plugins/tv.md
michielvandergeest: good point! will update

oscaruitenbroek: @wouterlucas why would we do this? Every app should be able to switch channels, there is no restriction needed for this.

PR #113, opened 2020-08-18
Feature/pin plugin

michielvandergeest:
Pin plugin for SDK with a default (simple) Pin dialog


wouterlucas (commented):
LGTM - only one minor comment is the different PINs certain systems can have. I know the simplest use case is 1 PIN to rule them all but I’ve seen more complex variations.

Does it make sense to specify which PIN we’re targeting and assume a default if the parameter is missing?

E.g. some of the “legacy” middlewares have a Purchase PIN, Profile PIN and Adult PIN on their system. All different lock states and granting access to different functionality on the system.

Idk if it makes sense for the first delivery of this, we could continue with a default for now and revise this later. But just dropping my thoughts here.

wouterlucas (approved):

PR #118, opened 2020-09-07
Feature/language

michielvandergeest:


docs/plugins/language.md
```diff
@@ -0,0 +1,207 @@
+# Language
+
+The Language plugin helps to make your App multilanguage.
```
wouterlucas, line 3: The language plugin supports internationalisation features for your application?

docs/plugins/language.md
```diff
@@ -0,0 +1,207 @@
...
+```js
+import { Language } from 'wpe-lightning-sdk'
+```
+
+### Loading on boot
+
+You can automatically load and initialize the Language plugin when your App boots, by specifying a `static` method `language()`
+on the `App`-class in `src/App.js`.
+
+The `language`-method should return which language to use initially. This can either be a string or a Promise that resolves a
```
wouterlucas, line 18: which language to use on initialisation.

docs/plugins/language.md
```diff
@@ -0,0 +1,207 @@
...
+    "introduction": "Hi, my name is {0}, my lastname is {1} and I am {2} years old",
+    "introduction2": "Hi, my name is {firstname}, my lastname is {lastname} and I am {age} years old",
+  }
+}
+```
+
+We can either pass a set of function argument for order based replacements.
+
+```js
+Language.translate('introduction1', 'John', 'Doe', 18)
```
wouterlucas, line 148: Niiiice.

docs/plugins/language.md
```diff
@@ -0,0 +1,207 @@
...
+Language.available() // ['en', 'nl', 'fr']
+```
+
+If you have specified an object with `names` in the `meta` key of your translations file, the `available`-method will return an
+Array of objects
+
+```js
+Language.available()
+// [{code: 'en', name: 'English', code: 'nl', name: 'Nederlands'}]
+```
```
wouterlucas, line 176: Mmmm, same function returning two different data structures. Are we sure about this?

wouterlucas (changes requested):

docs/plugins/language.md
michielvandergeest: @wouterlucas I see your point ... the alternative would be to always construct an array of objects (with `code` and `name`) and set name to equal the code, when no names are specified in the metadata. That would make the return consistent. Do you prefer that?

docs/plugins/language.md
michielvandergeest: changed

docs/plugins/language.md
michielvandergeest: 👍

PR #120, opened 2020-09-10
Added zipCode to Profile plugin.

michielvandergeest:
new zipCode property on profile plugin. Needed for the WeatherNetwork


wouterlucas (approved):
LGTM

Issue #131, opened 2020-10-19
Support Asynchronous Function Export from App's index.js


frank-weindel:
As a Lightning developer I'd like to load in my own remote configuration data (or do other things) in an asynchronous manner before initializing my app's App component.

Using the official SDK, you must return the result of a call to `Launch()` to the default exported function in your app's `index.js` file. Unfortunately the Launch result must be returned synchronously, so there is no ability to do something asynchronously before returning that value.

For example this is the default index.js file produced by `lng create`:
```
import { Launch } from '@lightningjs/sdk'
import App from './App.js'

export default function() {
  return Launch(App, ...arguments);
}
```

I'd instead like to transform that function into something more like this:
```
import { Launch } from '@lightningjs/sdk'
import App from './App.js'
import { loadConfig } from './config';
export default async function() {
  await loadConfig();
  return Launch(App, ...arguments)
}
```

That way my own asynchronous initializing and config loading can happen before the app is launched. Currently this pattern does not work since the caller of the function does not support a promise result.

frank-weindel:
@michielvandergeest @woutermeek I get that my PR was closed because of App platform work that would have to happen too. But is this something that is feasible? The workaround for modifying my `index.html` file for a self-hosted app (which mine is) is fine but does not work for my development with `lng dev`. Right now I'm stuck with `npm link`ing mt forked version of Lightning-SDK to achieve my ideal development work flow.

michielvandergeest:
As you say you can implement this functionality easily in the `index.html` created by `lng dist`, but the downside is indeed that you don't have the convenience of watching and automatically building (as you get with `lng dev`)

My first suggestion would be to add a custom watcher functionality to your own project. Could be something like this:

`npm install npm-watch --save-dev`

Then add this to your `package.json`:

```json
{
  "name": "myproject",
  "watch": {
    "dist": "src/**/*.js"
  },
  "scripts": {
    "dist": "lng dist",
    "watch": "npm-watch"
  }  
}
```

Finally run `npm run watch`. This will trigger the `lng dist` command everytime a je file changes in your `src` folder, similar to how `lng dev` does it.

Possibly we can add this functionality to the CLI as well. I'm adding it to the backlog in any case.

PR #132, opened 2020-10-19
Added ability to return a promise from an app's index.js function

frank-weindel:
Implements #131 

Tested locally both using the async promise approach and the original synchronous approach.

### DIFF
--- a/src/startApp.js
+++ b/src/startApp.js
@@ -52,13 +52,13 @@ const startApp = () => {
       console.time('app2')
       settings.appSettings.version = appMetadata.version
       settings.appSettings.id = appMetadata.identifier
-      app = window[appMetadata.id](
-        settings.appSettings,
-        settings.platformSettings,
-        settings.appData
-      )
-      canvas = app.stage.getCanvas()
-      document.body.appendChild(canvas)
+      return Promise.resolve(
+        window[appMetadata.id](settings.appSettings, settings.platformSettings, settings.appData)
+      ).then(_app => {
+        app = _app
+        canvas = app.stage.getCanvas()
+        document.body.appendChild(canvas)
+      })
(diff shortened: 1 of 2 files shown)


src/startApp.js
```diff
@@ -52,13 +52,13 @@ const startApp = () => {
...
-        settings.appSettings,
-        settings.platformSettings,
-        settings.appData
-      )
-      canvas = app.stage.getCanvas()
-      document.body.appendChild(canvas)
+      return Promise.resolve(
+        window[appMetadata.id](settings.appSettings, settings.platformSettings, settings.appData)
+      ).then(_app => {
+        app = _app
```
frank-weindel, line 58: Just an observation. I would have rather kept app locally scoped to avoid this line here. But there's a `finally` clause in the `getSettings` that relies on `app` existing. I'm not sure though _how_ `getSettings` could ever get a valid `app` object if it's being run BEFORE the app is retrieved from `window[appMetadata.id]`. Could you provide any clarity on that? Are those references to `app` required?

michielvandergeest: yes I see your point, it would be nicer to keep a local scope (and I believe that in an earlier version we actually did that),
but as it is now, the reference to the outer scope `app` is necessary

The `app` variable isn't actually accessed in the finally itself, but in an `onClose` method that is defined there as a platform setting. That `onClose`-callback is the one that ultimately needs a reference to the `app` in order to do some cleanup

michielvandergeest:
As discussed via Slack: startApp.js mimicks the Metrological App Platform during local development. So while this change works locally, it won't work in the context of the App store.

In the case of DIY hosting a likewise solution can be custom added to the `index.html` (created via `lng dist`-command).

PR #137, opened 2020-10-26
Fix/finally in startapp

michielvandergeest:
Fix for WRT-browser not properly handling finally. Only affects startApp.js (i.e. lng dev command) for live testing on STB.


wouterlucas (approved):
Alt Text

PR #138, opened 2020-10-27
feat(Language): get method

jhderojasUVa:
**Problem**

On SDK 2 Lang was a property called language that tells you the current language you are using. On SDK 3 there's not.

**Solution**

Added get method to Language to retrieve the current language you are using.


michielvandergeest (changes requested):
hi @jhderojasUVa thanks for your PR. I think this is good addition to the Language plugin.
Would you mind also including some short documentation on this functionality in your PR?
Then we should be good to go :) Thanks!

jhderojasUVa:
My fault to not update the doc! :)

michielvandergeest (approved):
LGTM

PR #141, opened 2020-10-30
Fix: wpe-lightning-sdk upgrade folder problem

jhderojasUVa:
**Problem**:
Sometimes when you are upgrading from SDK 2.x to SDK 3.x the postinstall.js script that runs after the installation fails to copy the files into some directories because doesn't know if it's running at wpe-lightning-sdk or at @lightningjs/sdk directory and need to have the correct path.

**Solution**:
Use and if where generating the node_modules directory that choose if you are in one place or another by looking the proccess.cwd includes @lightningjs or wpe-lightning-sdk (well, if you are in @lightningjs you are not on wpe-lightning-sdk :)

### DIFF
--- a/scripts/postinstall.js
+++ b/scripts/postinstall.js
@@ -23,10 +23,23 @@ const process = require('process')
 const replaceInFile = require('replace-in-file')
 const yesno = require('yesno')
 
-const nodeModulesFolder = path.join(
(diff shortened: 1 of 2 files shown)


michielvandergeest:
Closing PR, since alternative solution has been implemented in v3.1.1

PR #161, opened 2021-01-08
Patch 1

michielvandergeest:
@jlacivita creating a PR to make it easier to comment on the changes

### DIFF
--- a/src/Platform/index.js
+++ b/src/Platform/index.js
@@ -41,10 +41,10 @@ export const initPlatform = config => {
 
 const getOrSet = (key, params) => (params ? setInfo(key, params) : getInfo(key))
 
-const adSkipTiersEnum = {
+const adSkipTiersEnum = Object.freeze({
   noSkipNormalSpeed: 'NOSKIP_NORMAL_SPEED',
   allowSkipHighSpeed: 'ALLOW_SKIP_HIGH_SPEED'
-}
+})
 
 // public API
 export default {
--- a/src/Platform/index.js
+++ b/src/Platform/index.js
@@ -41,6 +41,11 @@ export const initPlatform = config => {
 
 const getOrSet = (key, params) => (params ? setInfo(key, params) : getInfo(key))
 
+const adSkipTiersEnum = {
+  noSkipNormalSpeed: 'NOSKIP_NORMAL_SPEED',
+  allowSkipHighSpeed: 'ALLOW_SKIP_HIGH_SPEED'
+}
+
 // public API
 export default {
   Localization: {
@@ -109,7 +114,7 @@ export default {
       return getOrSet('network', params)
     },
   },
-  Accesibility: {
+  Accessibility: {
     closedCaptions(params) {
       return getOrSet('closedCaptions', params)
     },
@@ -123,6 +128,18 @@ export default {
     },
     addToWatchHistory(history) {
       return getOrSet('watchHistory', history)
+    }
+  },
+  Advertising: {
+    AdSkipTiers: adSkipTiersEnum,
+    adSkipTier() {
+      return getOrSet('adSkipTier')
+    },
+    limitAdTracking() {
+      return getOrSet('limitAdTracking')
+    },
+    advertisingId() {
+      return getOrSet('advertisingId')
     }
   }
 }
--- a/src/Platform/index.js
+++ b/src/Platform/index.js
@@ -117,4 +117,12 @@ export default {
       return getOrSet('voiceGuidance', params)
     },
   },
+  Personalization: {
+    setEntitlements(entitlements) {
+      return getOrSet('entitlements', entitlements)
+    },
+    addToWatchHistory(history) {
+      return getOrSet('watchHistory', history)
+    }
+  }
 }


src/Platform/index.js
```diff
@@ -41,6 +41,11 @@ export const initPlatform = config => {
 
 const getOrSet = (key, params) => (params ? setInfo(key, params) : getInfo(key))
 
+const adSkipTiersEnum = Object.freeze({
+  noSkipNormalSpeed: 'NOSKIP_NORMAL_SPEED',
+  allowSkipHighSpeed: 'ALLOW_SKIP_HIGH_SPEED'
+})
+
```
michielvandergeest, line 48: I'm not sure I see the advantage of this ENUM technique (also the exporting it later on, so a dev can do `if (Platform.Advertising.adSkipTier === Platform.Advertising.AdSkipTiers.noSkipNormalSpeed) {}` versus just having (and documenting) the possible string values

src/Platform/index.js
```diff
@@ -109,12 +114,32 @@ export default {
...
+  Accessibility: {
     closedCaptions(params) {
       return getOrSet('closedCaptions', params)
     },
     voiceGuidance(params) {
       return getOrSet('voiceGuidance', params)
     },
   },
+  Personalization: {
+    setEntitlements(entitlements) {
```
michielvandergeest, line 126: I prefer we use `params` as the argument for the methods, to stay consistent with the other methods

Naming wise, to stay in line with the rest of the API, I think we should rename this method to simply `entitlements` (i.e `Platform.Personalization.entitlements(params?)`. I assume that in practice this API should be able to also retrieve the entitlements from the platform / transport layer? So the getter / setter pattern would make sense.

src/Platform/index.js
```diff
@@ -109,12 +114,32 @@ export default {
...
+      return getOrSet('entitlements', entitlements)
+    },
+    addToWatchHistory(history) {
```
michielvandergeest, line 129: Same here:

- `params` as the argument name
- consider to rename the method to simply `Platform.Personalization.watchHistory()` that functions as a getter / setter method.

src/Platform/index.js
```diff
@@ -109,12 +114,32 @@ export default {
...
+      return getOrSet('watchHistory', history)
+    }
+  },
+  Advertising: {
+    AdSkipTiers: adSkipTiersEnum,
```
michielvandergeest, line 134: See comment above, it seems to me the code for a developer becomes more verbose with this approach. What use case does it solve in practice?

src/Platform/index.js
```diff
@@ -109,12 +114,32 @@ export default {
       return getOrSet('network', params)
     },
   },
-  Accesibility: {
+  Accessibility: {
```
michielvandergeest, line 117: Good catch 🙈

michielvandergeest (commented):
See some code comments inline.

In general, I'm not sure about the separate namespacing of `Advertising` and `Personalization`.

Managing entitlements and watch history might deserve its own separate Plugin?

And things like `adSkipTier` seems like some kind of a user / device `Preference`? Maybe from a developer perspective it might make sense to introduce something like `Platform.Preferences.adSkipTier` along side other preference settings.

src/Platform/index.js
jlacivita: Hmmm. There's currently two separate features in the `Advertising` namespace:

- IFA-style anonymized ID for advertising. This allows app developers to integrate with our Ad Framework product.
- Ad Tier related methods, which is more of an entitlement / platform product than a preference.

I don't think they belong in Preferences, but i could be convinced we need a new place...

jlacivita:
> See some code comments inline.
> 
> In general, I'm not sure about the separate namespacing of `Advertising` and `Personalization`.
> 
> Managing entitlements and watch history might deserve its own separate Plugin?
> 
> And things like `adSkipTier` seems like some kind of a user / device `Preference`? Maybe from a developer perspective it might make sense to introduce something like `Platform.Preferences.adSkipTier` along side other preference settings.

Hmmm. There's currently two separate features in the `Advertising` namespace:

- IFA-style anonymized ID for advertising. This allows app developers to integrate with our Ad Framework product.
- Ad Tier related methods, which is more of an entitlement / platform product than a preference.

I don't think they belong in Preferences, but i could be convinced we need a new place...

src/Platform/index.js
jlacivita: The main value is code completion for developers

src/Platform/index.js
jlacivita: Yeah, I had exactly what you are suggesting initially.

I think we should have the conversation about methods that only go one way...

`setEntitlements` is set-only. you can't get them, since you're just passing them along to us so we know them. We're not the source of truth here, so it's not super useful to have a getter.

src/Platform/index.js
jlacivita: I have the same concern here. There's no way to get the watch history.

I tried to put verbs in each of these methods because they're not getter/setters, so a verb identifies what they do.

src/Platform/index.js
jlacivita: Code completion, and self documenting code.

src/Platform/index.js
jlacivita: nothing to see here :)

src/Platform/index.js
cletustboone: Does this technique actually have some effect in that regard? If so, I suppose that's a low-value add.

src/Platform/index.js
cletustboone: > _I think we should have the conversation about methods that only go one way..._

For something that is one-way, I wonder if it makes more sense to use an event emitter pattern that's fire and forget. Leave it up to the transport layer to listen and act on events. Does that pattern already exist in the SDK?

src/Platform/index.js
jlacivita: if the module exports the enum, then developers don't have to remember the various modes we come up with, and can just type `AdSkipTiers.` and see the options pop up.

Good developer experience is our first goal according to the requirements!

michielvandergeest:
Closing this PR since we've switched to separate smaller PRs for specific plugins (ie. Discovery, Authentication, Metrics).

PR #162, opened 2021-01-08
Migrating various Firebolt APIs into L-SDK

jlacivita:


bengreenberg:
I think It makes sense to have XVP stuff segmented, not sure about Ad Framework but maybe.

I dont think the name is right - there is Firebolt stuff elsewhere in the SDK as well so that seems confusing.

jlacivita:
Okay, I moved Advertising and Personalization back into Platform via a Profile submodule.

Also took a better (I think) attempt at matching L-SDK style w/ methods that have longer names.

src/Discovery/index.js
```diff
@@ -0,0 +1,68 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Log from '../Log'
+
+let getEntitlements = function() {
+  return {
```
bengreenberg, line 23: still think this should not return values - we don't have a way of getting them from XVP.

src/Discovery/index.js
```diff
@@ -0,0 +1,68 @@
...
+let setEntitlements = function() {
+  return true;
+}
+
+let watched = function(watchedItems) {
+  watchedItems.reduce( (items, item) => { Log.info('Added ' + watchedItems.watchedOn + ": " + watchedItems.contentId )})
+  return true
+}
+
+let dashBoardTile = function(name, imageUrl, linkUrl) {
```
bengreenberg, line 44: this is missing a bunch of parameters, probably should take an object. see https://ottx-docs.wc-g1.cf.comcast.net/docs/app-launchpad-tiles/

bengreenberg: also we need a better name for this....

src/Ads/index.js
```diff
@@ -16,223 +16,77 @@
...
+  return '26ccd5a7b2c2a50e7d4b2244e9d4c048'
 }
 
-const playSlot = (slot = []) => {
-  return slot.reduce((promise, ad) => {
-    return promise.then(() => {
-      return playAd(ad)
-    })
-  }, Promise.resolve(null))
+let clearAdvertisingId = function() {
```
bengreenberg, line 32: I think this is better called `resetAdvertisingId` as you still have an ad id at the end of it, just a new one.

src/Ads/index.js
```diff
@@ -16,223 +16,77 @@
...
-        cleanup()
-      },
-      // todo: pause, resume, mute, unmute beacons
-    }
-    // add all listeners
-    Object.keys(handlers).forEach(handler => videoEl.addEventListener(handler, handlers[handler]))
-
-    videoEl.play()
-  })
+let getConfig = function() {
```
bengreenberg, line 36: We may want to make this `getAdPlatformConfig` or something, as its to expected to be used generically,.

src/Ads/index.js
```diff
@@ -16,223 +16,77 @@
...
+    adSkipTier: 'NOSKIP_NORMAL_SPEED',		
+    adSkipGracePeriodSeconds: 60		
   }
 }
 
-const fireOnConsumer = (event, args) => {
-  if (consumer) {
-    consumer.fire('$ad' + event, args)
-    consumer.fire('$adEvent', event, args)
+let getPrivacy = function() {
```
bengreenberg, line 50: we need to work on the output format here, it needs to comply with some industry standards as badger does, I think.

src/Ads/index.js
```diff
@@ -16,223 +16,77 @@
...
 export const initAds = config => {
   if (config.getAds) {
-    getAds = config.getAds
+    initAdsPlayer(config)
   }
 }
 
-const state = {
-  active: false,
+let getAdvertisingId = function() {
```
bengreenberg, line 28: we also need to add methods:
`deviceAdAttributes()` (returns object)
`appStoreId()` (returns string)

src/Discovery/index.js
jlacivita: Well that sort of ends that discussion :) I'll remove

src/Discovery/index.js
jlacivita: Okay... renaming `watchNext` as discussed. will add the missing params

src/Ads/index.js
jlacivita: oh, good feedback

PR #163, opened 2021-01-11
Remove img deprecated methods

sandeep-vedam:
Following are the changes in this PR : 
1. Removed deprecated methods in the Image plugin source code
2. Removed deprecated methods from the Image plugin documentation
3. Updated Portrait method by removing the deprecated stuff under fix me

@michielvandergeest ,
Can you please closely look at point 3 above and take a call whether we need to remove it or keep as is?


michielvandergeest (approved):
LGTM

PR #169, opened 2021-01-12
(feat) Initial POC for Lifecycle & Events

wouterlucas:
Work in progress but ready for initial review.


jlacivita (changes requested):
I think this looks pretty good, although I think that `init` needs to be a method (not an event) that returns a `Promise`, so the app platform knows when the app is ready to be put in the `active` state.

Thoughts?

wouterlucas:
Depends on who puts the app in active state, the transport, the SDK or the app itself?

jlacivita:
I've been assuming that the transport layer does this stuff. which is why it'll be good start start syncing up on that :)

Here's a hot launch (app loaded before user requests it) and cold launch flow I wrote up before we started working with the L-SDK:

lifecycle_hot_launch
lifecycle_cold_launch

wouterlucas:
So in your flows its the app that's ultimately controlling the flow, that's possible. Though might have some negative side effects and puts more burden on the developer. Let's discuss.

jlacivita:
Hmm, the app controls resolving each promise, but the FEE initiates them.
The only thing the app initiates the `registerLifecycleHandler` which tells the FEE that this app cares about the different states.

Ideally, every app would have an`initialize()` (even if it's inherited from an abstract Application class)

wouterlucas:
Yes of course the event comes from a higher level with the exception of close being dual intent. However there is no reason for an app to stay in "init" state, it isn't a functional state. 

Suspend/resume however can have an implication. Close is something that needs to be handled carefully.

jlacivita:
Sounds like we're saying something similar? "init" isn't really a state, it's a method/routine/function.

I'd propose removing it from Event, and adding it to Application, e.g.:

```js
class MyApplication extends Lightning.Application {
    constructor(options) {
      const config = Deepmerge(defaultOptions, options)
      super(config)
      this.config = config
    }

    initialize() {
        _setup()
        return new Promise( (resolve, reject) => {
                // made this up, it'd be inside the 3rd party code
                callAppCustomAysnchronousInitializationCode().then( {} => { resolve() })
        })
    }
}

**Comment by wouterlucas** [comment]
That would make it lightning only and doesn't help. Technically that section already reports a ready through the metrics. 
If the discussion is promise vs event, it depends on what you want to do on the negative cases and whether or not there is a reason to do it synchronously.

**Comment by jlacivita** [comment]
Good point, we'd want to keep it decoupled from lightning.

If we define an interface, that all Firebolt apps require an `initialize` method w/ a specific signature, e.g. returns a `Promise`, then we can follow that signature in both Lightning.Application, as well as any broilerplate (non-lightning) sample Application that we provide.

**Comment by wouterlucas** [comment]
Absolutely, though what's left is events vs promises, let's discuss in the afternoon. I'd like to understand why a promise would be required.

**Comment by jlacivita** [comment]
👍

File: docs/plugins/lifecycle.md [hunk 1]
```diff
@@ -0,0 +1,66 @@
...
+
+The following states are available:
+
+| State         | Description                                 |
+| ------------- |:-------------------------------------------:|
+| init          | Application is initialising                 |
+| active        | Application is running, and in focus        |
+| background    | Application is running, not in focus        |
+| pause         | Application is paused, not in focus         |
+| close         | Application is closing and will be unloaded |
```
cletustboone, line 23: We have been referring to this as `unload` internally. Does `close` have precedent elsewhere in the SDK?

wouterlucas: We do have the concept of unloaded, which means the JS + DOM + graphics is unloaded, however the application itself will never have that as a status. I.e. you can't know you're dead yourself. The MAF Store app manager will know once an app is unloaded.

The close method has been in our SDK for a while and is transitional state. It means either someone externally or the app itself wants to close, this means the app will need to do its cleanup routine and once the close is done the upper layers can "unload" it.

jlacivita: makes sense to go with `close` to me, then.

docs/plugins/events.md
```diff
@@ -0,0 +1,45 @@
+# Events
+
+Generic Events emitter / listener interface which can be used for _any_ type of event. For Lifecycle events please see [Lifeycle](/plugins/lifecycle.md)
```
cletustboone, line 3: Should we allow developers to mix this in to their own objects?
```js
import { Events } from '@lightningjs/sdk'
const object = {}
const handler = (_args) => console.log('initialized')
object.mixin(Events)

object.addEventListener('initialize', handler)
```

src/Events/index.js
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+import Log from '../Log'
+
+const callbacks = {}
+
+const emit = (event, ...args) => {
+  callbacks[event] &&
+    callbacks[event].forEach(cb => {
+      cb.apply(null, args)
```
cletustboone, line 27: If we allow developers to mix this in to their own objects, then we should give them an option to provide a context for `cb.apply`.

docs/plugins/events.md
michielvandergeest: hey @cletustboone this looks like an interesting concept, but I'm not sure yet what problem it solves? Can you give an example of what `object` would be in a real App (as it's just an empty object literal now :) )

cletustboone: I was thinking about a component being able to do something with itself when something changes.
```js
someComponent.addEventListener('change:field', function() {someComponent.render()})
```
But then I remembered, there's already some existing mechanism in Lightning to accomplish this.

PR #171, opened 2021-01-15
Feature/discovery

jlacivita:
# Discovery
Your App likely wants to integrate with the Platform's discovery capabilities. For example to add a "Watch Next" tile that links to your app from platform home screen.

Getting access to this information requires to connect to lower level APIs made available by the platform. Since implementations differ between operators and platforms, the Lightning-SDK offers a Discovery plugin, that exposes a generic, agnostic interface to the developer.

Under the hood, an underlaying transport layer will then take care of calling the right APIs for the actual platform implementation that your App is running on.

The Discovery plugin is used to _send_ information to the Platform.

## Usage

Whenever you need to interact with the Discovery, import the Discovery plugin from the Lightning SDK

```js
import { Discovery } from '@lightningjs/sdk'
```


docs/plugins/discovery.md
```diff
@@ -0,0 +1,120 @@
+# Discovery
+Your App likely wants to integrate with the Platform's discovery capabilities. For example to add a "Watch Next" tile that links to your app from platform home screen.
+
+Getting access to this information requires to connect to lower level APIs made available by the platform. Since implementations differ between operators and platforms, the Lightning-SDK offers a Discovery plugin, that exposes a generic, agnostic interface to the developer.
+
+Under the hood, an underlaying transport layer will then take care of calling the right APIs for the actual platform implementation that your App is running on.
+
+The Discovery plugin is used to _send_ information to the Platform.
```
michielvandergeest, line 8: How does this relate to the `Discovery.entitlements()` method _getting_ entitlements?

docs/plugins/discovery.md
```diff
@@ -0,0 +1,120 @@
...
+```js
+import { Discovery } from '@lightningjs/sdk'
+```
+
+## Available methods
+
+### Entitlements
+Gets or sets the users entitlements for this app to enable platform home screen awareness.
+
+Takes an Array of entitlements, by default, during _local develompent_:
```
michielvandergeest, line 23: I'm not quite sure what you mean by _taking an array_ in this context? The defaults are generally used for return values during local development ...

jlacivita: Yeah, what I wrote sounds confusing. I'll rewrite

docs/plugins/discovery.md
jlacivita: The method only *sets* them. there's no way to get them.

jlacivita: ah, I see, the docs on line 4 were wrong. I fixed them.

PR #176, opened 2021-01-21
feat(media): loadstart, progress, rendition

jlacivita:
# Justification

Time-to-first-frame (TTF) and unexpected video stops are two of the biggest quality concerns of a video platform, and can lead to audience drop-off. Video quality is a close third, since users will not stick around for low bitrate video with distracting artifacts due to encode quality.

Our metrics API should make it easy to provide analytics on:

- how long it takes video to start up (TTFF)
- how often (and for how long) video is interrupted unexpectedly
- what bitrates end user are viewing over time

https://medium.com/pinterest-engineering/optimizing-video-playback-performance-caf55ce310d1

## Changes

- `Metrics.media.loadstart` - used in conjunction with `play` to calculate TTFF
- `Metrics.media.progress` - used to keep track of how far end users watch video (to see if other QoS improvements are working)
- `Metrics.media.rendition` - used to keep track of currently playing rendition, e.g. bitrate & dimensions
- Documentation of various Media events (based on HTML5 video events)

### DIFF
--- a/docs/plugins/metrics.md
+++ b/docs/plugins/metrics.md
@@ -221,28 +221,39 @@ The `Metrics.user` namespace is used for tracking media events.
 
 The [VideoPlayer plugin](/plugins/videoplayer) automatically tracks and sends Metrcs for the following media related events:
 
+- LoadStart
 - Abort
 - CanPlay
 - Ended
 - Pause
 - Play
+- Progress
 - Suspend
 - VolumeChange
 - Waiting
 - Seeking
 - Seeked
 - Ratechange
+- Rendition
 
 For apps that don't use the VideoPlayer plugin, tracking has to be implemented manually. Media metrics are initiated by
-calling the method `Metrics.media()` and passing it the url of the media asset that is being tracked.
+calling the method `Metrics.media()` and passing it the uri of the media asset that is being tracked. The uri may be a qualified URL to the CDN, or a unique identifier from a CMS.
 
 ```js
-const mediaMetrics = Metrics.media(assetUrl)
+const mediaMetrics = Metrics.media(assetUri)
 ```
 
 Next the following methods will become available on the `mediaMetrics` variable:
 
+#### Load Start
+Called when playback is requested, but before the frames are available.
+
+```js
+mediaMetrics.loadstart({ currentTime: ... })
+```
+
 #### Abort
+Playback was aborted, but not due to an error.
 
 ```js
 mediaMetrics.abort({ currentTime: ... })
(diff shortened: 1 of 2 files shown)


docs/plugins/metrics.md
```diff
@@ -255,61 +266,81 @@ mediaMetrics.canplay({ currentTime: ... })
...
 mediaMetrics.ratechange({ currentTime: ... })
 ```
 
+#### Rendition
+The playback rendition, e.g. bitrate, dimensions, etc., has changed.
+
+```js
+mediaMetrics.rendition({ currentTime: ..., bitrate: ..., width: ..., height: ... })
+```
+
```
michielvandergeest, line 343: Not 100% sure about the naming of this metric.

For reference, HLS.js calls this event 'level switching'. Shaka refers to it as an 'adaptation event'.
Would something like `qualitychange` fit here?

Any opinion on this @wouterlucas or @zombi6?

docs/plugins/metrics.md
```diff
@@ -221,28 +221,39 @@ The `Metrics.user` namespace is used for tracking media events.
...
 - VolumeChange
 - Waiting
 - Seeking
 - Seeked
 - Ratechange
+- Rendition
 
 For apps that don't use the VideoPlayer plugin, tracking has to be implemented manually. Media metrics are initiated by
-calling the method `Metrics.media()` and passing it the url of the media asset that is being tracked.
+calling the method `Metrics.media()` and passing it the uri of the media asset that is being tracked. The uri may be a qualified URL to the CDN, or a unique identifier from a CMS.
```
michielvandergeest, line 240: Yes, this is in line with what we discussed today. From an SDK perspective this will work fine (i.e. either a full url or an assetId as a grouping value).

@wouterlucas do you see any impact on the side of the stats server?

docs/plugins/metrics.md
jlacivita: rendition is pretty common language for both DASH & HLS:
HLS: https://tools.ietf.org/html/rfc8216
DASH: https://dashif.org/docs/workshop-2019/03-Reznik-DASH-IF-workshop-2019-CAE.pdf

Quality isn't right, because there are lots of attributes you can change across renditions, not just quality.

Adaptation works as well, but I see the word "rendition" in video context far more often.

michielvandergeest: Cool! I'm not completely familiar on the terminology for this, so just wanted to double check on using the correct naming. But I guess rendition makes sense.

Might want to change it to `renditionchange` (similar to the html5 standard `ratechange`)?

jlacivita: I'm cool with either `rendition` or `renditionChanged` although it might be nice if we could camel-case all of the multi-word methods in the API.

michielvandergeest: Great! `renditionchanged` it is then :)

On the subject of camelCasing: the current media metrics API is designed to be consistent with the html5 media events (which are all lowercase). This allows for direct (and more performant!) integration with the events emitted by the video tag.

I do agree that camelCase is more readable in general, so I've made a note to do a test next week to see whether we could support camelCase as well, without too much impact.

jlacivita: Cool, I'll make the "change" 😆.
We both mistyped in our last two comments... it's "change" (as you said originally) not "changed" assuming that's what we both meant for consistency.

jlacivita: oops, fixed.

docs/plugins/metrics.md
wouterlucas: Works for me.

wouterlucas (commented):
LGTM

PR #182, opened 2021-01-28
feat(lifecycle): added ready method, tweaked tense

jlacivita:
## Overview
Firebolt has a few use cases that require two-way communication around starting up an app.

As a platform owner, I want to show a loading screen, e.g. branding for the app, while the app is loading, and unable to show anything on it's own. I want to hide this loading screen once the app is "ready."

As a platform owner, I want to load some apps in the background to facilitate fast app launching, e.g. apps that are used frequently. I need a way for the app to let me know once it's minimally loaded, so I know when activating it will be a good experience for the user.
  
## Justification
Fast app launching is a high priority, but high resource feature. The platform needs a way to decide when it will be leveraged in order to best manage it's resources. Apps need to be prepared to be hot-launched (e.g. activated after previously being loaded into the background) or cold-launched (e.g. full load & activate all at once).

## Changes
- added `ready()` callback, that apps must call in order to be activated
- set initial state to "loading", once app calls `ready()`, platform can now decide whether to immediately activate the app, or put it in the background, pending a hot-launch later
- tweaked tense of 'paused' and 'closed'

### DIFF
--- a/src/Lifecycle/index.js
+++ b/src/Lifecycle/index.js
@@ -23,8 +23,8 @@ import isProbablyLightningComponent from '../helpers/isProbablyLightningComponen
 import { initBaseEvents } from './base'
 import { initLightningEvents } from './lightning'
 
-const supportedStates = ['init', 'active', 'pause', 'background', 'close']
-let currentState = 'init'
+const supportedStates = ['loading', 'active', 'paused', 'background', 'closed']
+let currentState = supportedStates[0]
 
 export const initLifecycle = (app, transport) => {
   if (transport) {
@@ -51,6 +51,7 @@ const stateChange = (state) => {
 
 // public API
 export default {
+  ready() { stateChange('background') },
   close() { stateChange('close') }, // when the app wants to close
   state() { return currentState },
   addEventListener,
--- a/src/Lifecycle/index.js
+++ b/src/Lifecycle/index.js
@@ -23,7 +23,7 @@ import isProbablyLightningComponent from '../helpers/isProbablyLightningComponen
 import { initBaseEvents } from './base'
 import { initLightningEvents } from './lightning'
 
-const supportedStates = ['loading', 'active', 'paused', 'background', 'closed']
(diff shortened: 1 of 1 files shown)


src/Lifecycle/index.js
```diff
@@ -23,8 +23,8 @@ import isProbablyLightningComponent from '../helpers/isProbablyLightningComponen
 import { initBaseEvents } from './base'
 import { initLightningEvents } from './lightning'
 
-const supportedStates = ['init', 'active', 'pause', 'background', 'close']
-let currentState = 'init'
+const supportedStates = ['loading', 'active', 'paused', 'background', 'closed']
```
wouterlucas, line 26: closed is a final state, the app will never know its closed. It can only have the intent of closing or an external request to close.

src/Lifecycle/index.js
```diff
@@ -51,6 +51,7 @@ const stateChange = (state) => {
 
 // public API
 export default {
+  ready() { stateChange('background') },
```
wouterlucas, line 54: why does ready result into a background state change?

src/Lifecycle/index.js
jlacivita: good point.

src/Lifecycle/index.js
jlacivita: Because it's up to the platform/transport layer to decide when it goes into active. might be a hot launch, might be a cold launch.

wouterlucas: but this is not the transport layer, this is the SDK. By doing the above you hardcode the ready() api to always set the state to background, effectively nullifying the usage of the ready() API.

Agree this should be done in the transport layer it, not in here. These are just the APIs and they should map accordingly, e.g. ready should set ready as a state.

jlacivita: agreed, I let a specific feature leak into the SDK defaults. will fix.

jlacivita:
Closing in favor of https://github.com/rdkcentral/Lightning-SDK/pull/183

PR #183, opened 2021-02-01
Feature/lifecycle and events

michielvandergeest:


docs/plugins/advertising.md
```diff
@@ -0,0 +1,165 @@
+# Advertising
+
+The _Advertising_ plugin helps to enable (video) advertisements in an App.
+
+The plugin can be used as a drop-in solution that takes care of the entire Ad playback flow (in combination with the VideoPlayer plugin),
+as well as the integration with different advertisment distributers depending on the specific platform deployment.
```
cletustboone, line 6: distributors (sp)

docs/plugins/advertising.md
```diff
@@ -0,0 +1,165 @@
...
+### Customize Ad retrieval
+
+In some cases you might want to leverage the built in Ad playback logic, but you might require some customizations to the way that ads are retrieved from the ad distributor.
+
+For this usecase you can overwrite the `getAds` method in the Advertising plugin and supply it with your own. The `getAds` function should return a `Promise`
+that resolves an object with advertisements.
+
+```js
+Advertising.getAds = function() {
+  return Promise.resolve({
```
cletustboone, line 81: Is the shape of this advertisement object prescribed somewhere?

docs/plugins/events.md
```diff
@@ -1,45 +1,138 @@
...
+
+Events.listen('Platform', 'device.hdcp', val => {
+  console.log('The hdcp value changed on the platform!', val)
+})
+```
+
+2) Listen to all events of a specific Plugin
+
+```js
+Events.listen('Lifecycle', (val, plugin, event) => {
```
cletustboone, line 42: `plugin` in the callback signature is redundant here, isn't it?

docs/plugins/events.md
```diff
@@ -1,45 +1,138 @@
...
 
-### Remove event listener
+### Clear
+
+Registered events can be cleared
+
+1) By id
+
+```js
+const listenerId = Events.listen('Platform', 'device.dhcp', () => {})
```
cletustboone, line 84: Providing an empty callback to indicate "I'm not listening anymore!" is odd to me. Is there any reason not to coat this with a little sugar like `Events.clear('Platform', 'device.hdcp')`?

cletustboone: Oh sorry, I should have scrolled down some more :)

docs/plugins/lifecycle.md
```diff
@@ -17,9 +20,10 @@ The following states are available:
 | State         | Description                                 |
 | ------------- |:-------------------------------------------:|
 | init          | Application is initialising                 |
+| ready         | Application is ready for user interaction   |
 | active        | Application is running, and in focus        |
 | background    | Application is running, not in focus        |
-| pause         | Application is paused, not in focus         |
+| pause         | Application is pausing, not in focus        |
```
cletustboone, line 26: I think I'd prefer these to be called `pausing` and `closing` so that it's clear that these are transitional. I suppose the same goes for `initializing`.

jlacivita:
This is a really robust event API, I like it!

A couple minor notes in the review, and one overall note is around case sensitivity around plugin names and sub-module names, e.g. `Platform.Device.xxx`.

Metrics uses lower case sub-modules, while other packages use upper case, and your samples in events.md use lower-case.

I think we should consider making plugin and sub-module names case insensitive, and probably event names, too.

docs/plugins/advertising.md
```diff
@@ -0,0 +1,165 @@
...
+import { Advertising } from '@lightningjs/sdk'
+```
+
+## Drop-in Advertising solution
+
+The drop-in usecase allows you to setup advertisements for your App with minimal code and configuration.
+
+### Setup
+
+In order to enable ads, both the _VideoPlayer_ and _Advertising_ plugin need to be imported. Next the _Advertising_ plugin
```
bengreenberg, line 26: Are we guaranteeing the the new Ad Router-based Ad Framework will work fully with the Lightning SDK player?

docs/plugins/advertising.md
```diff
@@ -0,0 +1,165 @@
...
+
+### AdvertisingId
+
+Retrieves the Advertising Id from the _Platform transport layer_. Returns a `Promise`, that resolves an `String` with the id.
+When the method is passed `true` as an argument a new Advertisement Id will be generated.
+
+
+```js
+Advertising.advertisingId()
+Advertising.advertisingId(true) // regenerate the id
```
bengreenberg, line 141: I'm not sure ID regeneration should be allowed via this API. It will be a distributor-only method. Additionally we refer to this as a *reset*.

docs/plugins/advertising.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+During _local development_ the advertisingId will be randomly generated.
```
bengreenberg, line 144: It may be better to return all zeroes to more explicitly show that this is a bogus value.

docs/plugins/discovery.md
```diff
@@ -0,0 +1,122 @@
...
+Sets the users entitlements for this app to enable platform home screen awareness.
+
+Note that this overrides previous entitlements sent to this method.
+
+Takes an Array of entitlements:
+
+```js
+Discovery.entitlements([
+  {
+    entitlementId: "http://entitlements/some/canonical/id",
```
bengreenberg, line 30: XVP entitlements will not be URIs so this example could be made more realistic.

docs/plugins/discovery.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Watched
+Adds a content the user's watch history to enable platform home screen awareness.
+
+Takes an array of watched items:
+
+```json
+Discovery.watched([
+  {
+    "contentId": "http://content/some/canonical/id",
```
bengreenberg, line 50: same comment, content IDs are not URIs.

jlacivita:
## Overview
Moved this here for context.

Firebolt has a few use cases that require two-way communication around starting up an app, as well as general use cases for change notifications.

As a platform owner, I want to show a loading screen, e.g. branding for the app, while the app is loading, and unable to show anything on it's own. I want to hide this loading screen once the app is "ready."

As a platform owner, I want to load some apps in the background to facilitate fast app launching, e.g. apps that are used frequently. I need a way for the app to let me know once it's minimally loaded, so I know when activating it will be a good experience for the user.

As an app developer, I want to be made aware of changes to lifecycle states, as well as other properties, e.g. Platform status and settings.
  
## Justification
Fast app launching is a high priority, but high resource feature. The platform needs a way to decide when it will be leveraged in order to best manage it's resources. Apps need to be prepared to be hot-launched (e.g. activated after previously being loaded into the background) or cold-launched (e.g. full load & activate all at once).

Change notification for settings and state is fundamental the evolution of the platform.

jlacivita (changes requested):

docs/plugins/advertising.md
michielvandergeest: good one, fixed :)

docs/plugins/advertising.md
michielvandergeest: No, it's not described sufficient enough atm. It's on the todo list and might need some further hashing out.

docs/plugins/events.md
michielvandergeest: yup .. that was bugging me a bit as well :)

ideally I'd only pass non-redundant arguments, but not sure yet if I can pull that off without too much extra code. I've been considering simply changing the order to `(val, event, plugin)`, which might be just what we need

docs/plugins/events.md
michielvandergeest: Yup, this was just a quick and dirty example. The callback in the listener is just a void function. And it's clearing it by id.
Agreed that in the final docs that should be a more clear example.

docs/plugins/lifecycle.md
michielvandergeest: yes, I guess that's a valid point .. @wouterlucas what do you think?

docs/plugins/events.md
```diff
@@ -1,45 +1,138 @@
...
+Events.clear('Platform') // removes all listeners to the Platform plugin
 ```
 
-### Emit events
+3) By Plugin and Event
+
+```js
+Events.listen('Platform', 'device.dhcp', () => {})
+Events.listen('Platform', 'localization.language', () => {})
+Events.listen('Lifecycle', 'init', () => {})
```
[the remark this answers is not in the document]
michielvandergeest: yes, in the quick and dirty docs I've used empty void callbacks and that can cause some confusion (both you and  @cletustboone got thrown off). Will be improved in the final documentation.

docs/plugins/lifecycle.md
```diff
@@ -17,9 +20,10 @@ The following states are available:
 | State         | Description                                 |
 | ------------- |:-------------------------------------------:|
 | init          | Application is initialising                 |
+| ready         | Application is ready for user interaction   |
```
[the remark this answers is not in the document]
michielvandergeest: I understand the transition should be instant or quick, making the ready state a (very) short lived state. So from an App developer perspective it's probably not the most useful state to listen to.

Is there anything we can think of that would vouch against explicitly passing through the `ready` state?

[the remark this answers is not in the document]
jlacivita: Hmm..

**Cons** of having `ready` as an explicit state:

1. keeping the number of states as few as possible
2. the app itself initiates this state, so it already knows
3. Anything it might want to do internally on `ready` should have been done before `ready()` was called
4. developers might wrongly assume that either `active` or `background` is specifically next and do the wrong thing during the `ready` listener.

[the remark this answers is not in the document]
michielvandergeest: fair enough .. then let's remove the ready state all together, and just keep the `ready()` method as a one-way street to inform the platform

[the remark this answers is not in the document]
jlacivita: 👍

docs/plugins/lifecycle.md
jlacivita: What do you mean by transitional, here, @cletusboone? Aren't all state machines are transitional?

docs/plugins/events.md
jlacivita: it's not redundant if a developer creates a function that's going to handle multiple events from different plugins, and just pass a reference to that function into `listen`. You'd need a way to fully distinguish the event, which is `plugin + event`.

maybe for the callback, we need a single object, with fields, e.g.:

```javascript
{
    value: 'foo',
    module: 'Platform',
    event: 'device.hdcp'
}
```

jlacivita (changes requested):

michielvandergeest:
> This is a really robust event API, I like it!
> 
> A couple minor notes in the review, and one overall note is around case sensitivity around plugin names and sub-module names, e.g. `Platform.Device.xxx`.
> 
> Metrics uses lower case sub-modules, while other packages use upper case, and your samples in events.md use lower-case.
> 
> I think we should consider making @plugin and sub-module names case insensitive, and probably event names, too.

@jlacivita regarding this comment: yes, case sensitivity is important and should be consistent across the SDK.

In retrospect and looking at the existing API, I think that the "namespaces" in the `Platform` plugin should have been all lowercase. So I suggest we simply change accordingly in the Platform plugin (i.e. `Platform.Device.dhcp` => `Platform.device.dhcp`).

jlacivita:
> @jlacivita regarding this comment: yes, case sensitivity is important and should be consistent across the SDK.
> 
> In retrospect and looking at the existing API, I think that the "namespaces" in the `Platform` plugin should have been all lowercase. So I suggest we simply change accordingly in the Platform plugin (i.e. `Platform.Device.dhcp` => `Platform.device.dhcp`).

Makes sense to me, and it seems like "Metrics" was the only module in the main branch w/ namespaces, so that'll also be consistent w/ what folks are doing in production.

Let's address it separately in a different PR.

docs/plugins/events.md
cletustboone: I'd avoid returning objects in the absence of type declarations. As a user, it's easier for me to grok the callback signature. Plus, I _think_ you'll get better IDE behavior out of the box with a sane callback signature as opposed to an opaque object without type decs.

docs/plugins/lifecycle.md
cletustboone: I guess they are, but there's a subtle difference in the wording for each state. Notice that `active` says that the "Application **is** running," but the `close` state says that it _is closing_.

The difference being that the app will receive a signal from the control plane to close itself, and the app might execute logic. It will not know that it is closed. It can only know that it _should_ close and what it can tell the control plane is that it _is closing._ So if you query the state at that point, we might as well be semantically correct about what's actually going on.

For `pause`, maybe it doesn't matter since the app can presumably know that it is paused or that it _will pause_ and it can report this state to the control plane.

docs/plugins/events.md
michielvandergeest: I agree completely with you @cletustboone ..

I've refactored this part in this commit: https://github.com/rdkcentral/Lightning-SDK/commit/020a04a23768d9eaddabc92121929117ccd03aaf

I think this API looks nicer and makes more sense from an App dev perspective.

jlacivita: I do like the way this looks w/ the listener functions declared right inside of the `listen()` call, but what if someone wants to reuse an existing function, e.g.:

```javascript
function myCustomListener(plugin, event, value) {
  if (plugin === 'Platform') {
    if (event === 'device.hdcp') {
      ...
    }
    else if (event === 'accessibility.closedCaptionsToggled') {
      ...
    }
  }
}

Events.listen('Platform', 'device.hdcp', myCustomListener)
Events.listen('Platform', 'accessibility.closedCaptionsToggled', myCustomListener)
```

in that case, you need all three variables passed to each version of `listen` to get it to work.

Listening to all events seems like it might not be the most efficient way around this.

docs/plugins/events.md
```diff
@@ -0,0 +1,165 @@
...
+
+```js
+Events.listen('Lifecycle', (event, value) => {
+  if(event === 'init') {
+    // execute init functionality
+  }
+  if(event === 'close') {
+    // execute close functionality
+  }
+})
```
jlacivita, line 51: I do like the way this looks w/ the listener functions declared right inside of the `listen()` call, but what if someone wants to reuse an existing function, e.g.:

```javascript
function myCustomListener(plugin, event, value) {
  if (plugin === 'Platform') {
    if (event === 'device.hdcp') {
      ...
    }
    else if (event === 'accessibility.closedCaptionsToggled') {
      ...
    }
  }
}

Events.listen('Platform', 'device.hdcp', myCustomListener)
Events.listen('Platform', 'accessibility.closedCaptionsToggled', myCustomListener)
```

in that case, you need all three variables passed to each version of `listen` to get it to work.

Listening to all events seems like it might not be the most efficient way around this.

docs/plugins/lifecycle.md
```diff
@@ -0,0 +1,116 @@
...
+```
+
+## Available states
+
+The following states are available:
+
+| State         | Description                                 |
+| ------------- |:-------------------------------------------:|
+| init          | Application is initialising                 |
+| ready         | Application is ready to be displayed        |
```
jlacivita, line 23: Can we change:

Application is ready to be displayed

to

Application is ready to be displayed and pushed into the either the `active` or `background` state depending on whether it was launched by the user, or was pre-loaded by the platform.

jlacivita: This came up as a confusion point today

jlacivita (changes requested):

docs/plugins/lifecycle.md
wouterlucas: I think that's a little too much text for the table. Maybe we need a seperate section in the documentation that describes the lifecycle in its entirety (and not just the APIs)?

wouterlucas (approved):

PR #190, opened 2021-02-17
Soft exit vs hard exit

jlacivita:
## Overview
The Lifecycle API needs a way for applications to do both soft exits and hard exits.

A soft exit moves the app to the `background` state, and returns the user to the home screen.

A hard exit moves the app through the `closing`, and `closed` states.
  
## Justification
This is to enable two use cases:

- As an app developer, I want my application to go in the `paused` state when the user hits _back_ on their remote, in case the user decides to come back quickly, or left the app accidentally.

- As an app developer, I want my application to go into the `closing` and then `closed` states when the user manually selects an "Exit" option from the menu, so that when they return the app is reloaded to the home screen.

## Changes
- Propose we rename `close()` to `exit()`
- Add an optional `unload` (boolean) parameter to exit

### DIFF
--- a/src/Advertising/index.js
+++ b/src/Advertising/index.js
@@ -8,11 +8,8 @@ export default {
   policy() {
     return Platform.get('advertising.policy')
   },
-  advertisingId(reset = false) {
-    return reset
-      ? Platform.get('advertising.advertisingId')
-      : // 'set null' means 'recreate new id'
-        Platform.set('advertising.advertisingId', null)
+  advertisingId() {
+    return Platform.get('advertising.advertisingId')
   },
   deviceAttributes() {
     return Platform.get('advertising.deviceAttributes')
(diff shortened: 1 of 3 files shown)


wouterlucas:
I dont think it is up to the application to decide whether or not to go into background state. That choice ultimately lives at the resident application and that's where entire business logic to background vs close logic should lie.

When the user presses menu the menu key isn't handled by the application. In fact the menu key shouldn't even be provided to the application to begin with. Instead the RA will handle the menu key as a set of reserved keys and decide whether or not the application should be closed or put in the background (dependent on the type of device, resources available and whatever other variables that are solely known to the RA). 

The other case where the user exits out of the application is a little more interesting. Still that decision, whether or not to background, should lie with the Resident Application and not the 3rd party app. To keep it simple a user oriented exit can be handled as an exist, the user did decide to back out / click exit v.s. press the menu key. Still the RA could decide, and we should check if that's possible, to not fully close the app but on closing put into paused instead. 

However in both cases the exit flow from the application perspective is the same:
* Menu key case isn't handled by the application
* Exit should always be the same flow regardless of whether or not the application _actually_ gets closed or put in the back ground
* The decision to background lies with the resident application.

PR #192, opened 2021-02-18
Feature/metadata

sandeep-vedam:
Added Metadata Plugin . Following are the methods that are added in this PR : 
1. get - to retrieve the value of any item related to metadata .
2. appId - to retrieve app id from metadata
3. appName - to retrieve app name from metadata
4. appVersion - to retrieve app version from metadata without any hashes
5. appIcon - to retrieve app icon from metadata
6. appFullVersion - to retrieve the app full version from metadata


src/Metadata/index.js
```diff
@@ -0,0 +1,49 @@
...
+let metadata = {}
+
+export const initMetadata = metadataObj => {
+  metadata = metadataObj
+}
+
+export default {
+  get(key, fallback = undefined) {
+    const val = metadata[key]
+    return val !== undefined ? val : fallback
```
michielvandergeest, line 28: I think it's better to write it like this: `return key in metadata ? metadata[key] : fallback`

src/Metadata/index.js
```diff
@@ -0,0 +1,49 @@
...
+  },
+  appId() {
+    return this.get('id')
+  },
```
michielvandergeest, line 32: Can we also add a `safeAppId` (the one that's also used in startapp.js)?

src/Metadata/index.js
```diff
@@ -0,0 +1,49 @@
...
+  appName() {
+    return this.get('name')
+  },
+  appVersion() {
+    let ver = this.get('version')
```
michielvandergeest, line 37: do we need this intermediate variable?

src/Metadata/index.js
```diff
@@ -0,0 +1,49 @@
...
+    // Get only version and not the hashvalue
+    let version = ver.split('-')
```
michielvandergeest, line 39: if (for some reason) the version is undefined (the default fallback of get) calling split will yield an Error.

michielvandergeest (changes requested):

src/startApp.js
```diff
@@ -49,15 +49,19 @@ const startApp = () => {
...
+      let bundle = window[appMetadata.safeId]
       // support rollup and esbuild
       if (typeof bundle !== 'function') {
         bundle = bundle.default
       }
 
       console.time('app2')
-      settings.appSettings.version = appMetadata.version
+      //Adding complete metadata info to app settings
+      settings.appSettings = { ...appMetadata }
```
michielvandergeest, line 60: I think if we use a simple `Object.assign` it will yield the same effect, without generating more code in the compiled (es5) version. Can you check this?

src/Metadata/index.js
```diff
@@ -0,0 +1,48 @@
...
+
+export default {
+  get(key, fallback = undefined) {
+    return key in metadata ? metadata[key] : fallback
+  },
+  appId() {
+    return this.get('id')
+  },
+  safeAppId() {
+    return this.get('safeId')
```
michielvandergeest, line 33: we can't actually return safeId directly (because it wouldn't work in the store). Instead we should repeat the same regex replacement as we use in startApp.js here in the plugin. Can you add that @sandeep-vedam ?

michielvandergeest (changes requested):

michielvandergeest (approved):

PR #194, opened 2021-02-19
Feature/docs

LBakkerTDS:
First correction of first WIP Review (in fpscounter) to see if push from Git Desktop works correctly.


docs/plugins/image.md
```diff
@@ -1,77 +1,81 @@
 # Image
 
-The standard way of displaying images in Lightning is to just specify the `src`. This is the prefered way for your App's local assets (such as background, splash screen, logo and icons).
+The standard way of displaying images in Lightning is to just specify the `src`. This is the preferred way for your local assets (such as background, splash screen, logo and icons) of Lightning Apps.
```
michielvandergeest, line 3: no need for "your" here I think (your local assets => local assets)

docs/plugins/image.md
```diff
@@ -1,77 +1,81 @@
...
 
 ```js
 import { Img } from '@lightningjs/sdk'
 ```
 
-## Available methods
+## Available Methods
 
-### Exact
+### exact()
```
michielvandergeest, line 19: I'm not sure if I like this change ... (multiple occurrences, so only mentioning once)
Let's discuss

docs/plugins/image.md
```diff
@@ -1,77 +1,81 @@
...
 
 ```js
 Img(url).portrait(height)
 ```
 
-### Cover
+### cover()
 
-Resizes the image in such a way that it covers the entire area. Depending on the orientation (portrait or landscape) of the image it will resize the image by width or by height.
+Resizes the image in such a way that it covers the entire area. Depending on the orientation (portrait or landscape) of the image and the target, it resizes the image by width or by height.
```
michielvandergeest, line 46: not sure if 'target' is clear enough, maybe should be something like "Depending on the orientation (portrait or landscape) of the source image and the orientation of the final result"

docs/plugins/image.md
```diff
@@ -1,77 +1,81 @@
...
+Generates the image without resizing it (that is, it uses the original dimensions), while still passing it through the proxy (and taking advantage of caching).
 
 ```js
 Img(url).original()
 ```
 
-## Image quality
+## Image Quality
+
+To increase the performance on lower-end boxes –especially those with limited GPU memory– there is a platform setting `image.quality` which enables you to control the *image quality*.
```
michielvandergeest, line 70: I think platform should be stressed (italic), to indicate it's not an application setting.

michielvandergeest (changes requested):
WIP review

docs/plugins/image.md
LBakkerTDS: Agreed.

docs/plugins/image.md
LBakkerTDS: Agreed to remove the parentheses (see discusion in Skype chat of February 23rd).

docs/plugins/image.md
LBakkerTDS: Suggestion: "Depending on the orientation (...) of the source image and that of the final result, it [...]."

docs/plugins/image.md
LBakkerTDS: I'm not sure if the italic emphasis indicates that it's not an application setting. Isn't there another way of pointing this out? E.g., "[...] there is an _external_ platform setting 'image.quality' which [...]"?

docs/plugins/image.md
michielvandergeest: maybe we should say "the desired output" instead of "the final result" ?

docs/plugins/image.md
michielvandergeest: I think using "external" could add more confusion actually (cause we don't mention that anywhere else). The importance is to distinguish that image quality is placed in the 'platform' section of the Settings. Perhaps linking to the appropriate docs on settings will be enough?

docs/plugins/image.md
LBakkerTDS: You suggest the following: "Depending on the orientation (...) of the source image and that of the desired output, it [...]."

I think 'desired output' instead of 'final result' is fine, since it covers the meaning of the 'result' in a more appropriate manner. Note: this will be adjusted for both the 'cover' and 'contain' methods.

docs/plugins/image.md
LBakkerTDS: I'll include a link to [http://www.lightningjs.io/docs/#/lightning-sdk-reference/plugins/settings?id=settings](http://www.lightningjs.io/docs/#/lightning-sdk-reference/plugins/settings?id=settings), where the Application and Platform Settings stored in **settings.json** are introduced.

docs/plugins/profile.md
```diff
@@ -1,150 +1,151 @@
...
 ```js
 import { Profile } from '@lightningjs/sdk'
 ```
 
-## Available methods
+## Available Methods
 
-When you call a method without params it will _return_ the Profile information in the form of a promise.
-If you pass params, it will _update_ the Profile information.
+If you call one of the methods below *without* parameters, it returns the Profile information in the form of a *promise*. If you pass parameters, the method *updates* the Profile information.
```
michielvandergeest, line 19: important detail that I think we should add: if you pass params the method will _attempt_ to update (the operator / platform decide whether something actually can be updated)

docs/plugins/purchase.md
```diff
@@ -1,31 +1,33 @@
...
-- fires a _payment request_ to the Metrological Billing server
-- attempt to _register a successful purchase_ at the CSP backend
-- _confirms_ back to the Metrological Billing that the purchase has been registered by the CSP backend
+In a typical, default setup, buying an asset with the Lightning SDK and the Metrological Billing Server only requires calling `Purchase.buy()`. No other calls should be necessary.
 
-Using this method requires that the Purchase plugin is properly _configured_ via the `Purchase.setup()` method.
+The `buy` method performs the following actions:
 
-The `buy`-method accepts an assetId (as defined by the CSP) as it's first argument.
+* Retrieve a signature for a gives asset
```
michielvandergeest, line 25: given

docs/plugins/profile.md
```diff
@@ -1,150 +1,151 @@
...
 
 ```js
 Profile.household()
 ```
 
-### Language
+### language
 
-Gets the language. During _local development_ will attempt to return the browser's language, with a default to `en`.
+Gets the user's mother language. During *local development*, it attempts to return the browser's language with `en` as default.
```
michielvandergeest, line 71: not sure if mother language is appropriate here?

docs/plugins/purchase.md
```diff
@@ -36,22 +38,16 @@ Purchase.buy(assetId).then(() => {
...
 
-The method accepts a config `object` as its only argument. This object can have 2 keys: `cspUrl` and `endpoints`.
+> The `setup` method should only be called once, before any other Purchase plugin method is called.
 
-`cspUrl` specifies the _base Url_ of the backend of the CSP.
+The `setup` method accepts a *configuration object* as its only argument. The object can have *2 keys*:
 
-`endpoints` can be used to customize those API endpoints that differ from the default specification (as outlined below). For each _endpoint_ that deviates from the default, specify the _uri_ (which will be appended to the `cspUrl`, unless specified as a fully qualified domain) and the REST _method_ to use.
-
-If needed you can also specify an object with custom `headers` that will be merged in with the defaults headers (`{ Accept: 'application/json', 'Content-Type': 'application/json' }`)
```
michielvandergeest, line 54: I think this sentence was maybe added at a later stage. We need to add this info as well.

docs/plugins/purchase.md
```diff
@@ -88,72 +84,32 @@ Purchase.setup('http://csp-backend.com/api', {
...
     method: 'DELETE',
   },
-  signature: {
-    uri: '/assets/:id/signature',
-    method: 'POST',
-    headers: {
-      'x-auth-token': 'abcd123xyz'
-      'cache-control': 'no-cache'
-    }
-  },
```
michielvandergeest, line 98: I think this example was added afterwards. We should re-add it.

docs/plugins/purchase.md
```diff
@@ -88,72 +84,32 @@ Purchase.setup('http://csp-backend.com/api', {
...
 })
 ```
 
-In some cases the CSP backend setup is too different from the default specification, that it becomes too difficult
-to configure the enpoint uri's. In those cases you can specify a `callback` instead, and handle the request to the
-CSP backend entirely yourself.
+> If the CSP backend does *not* completely follow the expected blueprint, or can not be configured properly via the `setup` method, the steps in the `buy` method can also be implemented *manually*. In that case, you can use the Plugin methods `signature`, `payment`, `subscribe` and `confirm` (see [below](#signature)).
 
-The callback function will receive 2 arguments: `params` and `data` and should _always_ return a `Promise` which resolves the result (or rejects an error in case of a failure). When a callback is specified, all other keys in that `endpoint` config
-are ignored.
```
michielvandergeest, line 107: Info about callbacks might have been added afterwards and got lost now

docs/plugins/purchase.md
```diff
@@ -88,72 +84,32 @@ Purchase.setup('http://csp-backend.com/api', {
...
-```
-
-### Assets
-
-The `assets` method returns a Promise with the result of the request to the CSP to retrieve assets available for purchase. In most cases it would return an `Array` of objects with information about the assets (i.e. _title_, _description_ and _price_).
-
-The `household` is automatically sent as a `query param` in the request, allowing to optionally return information on whether the asset has been _purchased_ by the current household.
-
-The `assets` method is an _optional_ convenience method exposed by the Purchase plugin. Depending on the CSPs API integration, the functionality to retrieve assets can be implemented fully custom in an App.
+You can use the `household` option to return information on whether the assets were purchased by the current household or not. It is automatically added as a `query param` in the request.
```
michielvandergeest, line 96: I think the previous sentence was more clear ..

docs/plugins/purchase.md
```diff
@@ -88,72 +84,32 @@ Purchase.setup('http://csp-backend.com/api', {
...
 })
 ```
 
-### Asset
+### asset
 
 The `asset` method returns a Promise with the result of the request to the CSP to retrieve details for a specific asset.
 
-The `household` is automatically sent as a `query param` in the request, allowing to return information on whether the asset has been _purchased_ by the current household.
+You can use the  `household` option to return information on whether the asset was purchased by the current household or not. It is automatically added as a `query param` in the request.
```
michielvandergeest, line 110: same comment as before. I think the new sentence isn't more clear than the old one

docs/plugins/purchase.md
```diff
@@ -88,72 +84,32 @@ Purchase.setup('http://csp-backend.com/api', {
...
-### Assets
-
-The `assets` method returns a Promise with the result of the request to the CSP to retrieve assets available for purchase. In most cases it would return an `Array` of objects with information about the assets (i.e. _title_, _description_ and _price_).
-
-The `household` is automatically sent as a `query param` in the request, allowing to optionally return information on whether the asset has been _purchased_ by the current household.
-
-The `assets` method is an _optional_ convenience method exposed by the Purchase plugin. Depending on the CSPs API integration, the functionality to retrieve assets can be implemented fully custom in an App.
+You can use the `household` option to return information on whether the assets were purchased by the current household or not. It is automatically added as a `query param` in the request.
 
+The `assets` method is an *optional* convenience method provided by the Purchase plugin. Depending on the CSP's API integration, you can implement the functionality to retrieve assets in your App in a fully customized manner.
```
michielvandergeest, line 98: shouldn't it be "custom manner"?

docs/plugins/purchase.md
```diff
@@ -162,13 +118,15 @@ Purchase.asset(assetId).then(assets => {
...
-### Signature
+### signature
+
+To make a payment request to the Metrological Billing Server, the CSP must provide a *signed signature*.
 
-In order to make a payment request to the Metrological Billing server, the CSP needs to provide a signed signature.
+The `signature` method returns a *Promise* with the result of the request made to the CSP to retrieve that signature. It accepts the `id` of the asset to purchase as its argument.
 
-The `signature` method returns a Promise with the result of the request to the CSP to retrieve that signature. The `signature`-method accepts the `id` of the asset to purchase as it's argument. The `household` will be automatically be added  as a `query param` in the request.
+The `household` option is automatically added  as a `query param` in the request.
```
michielvandergeest, line 127: household value

michielvandergeest (changes requested):

docs/plugins/language.md
```diff
@@ -1,45 +1,44 @@
...
 
-### Loading on boot
+### Loading on Boot
 
 You can automatically load and initialize the Language plugin when your App boots, by specifying a `static` method `language()`
-on the `App`-class in `src/App.js`.
+on the `App` class in **src/App.js**.
 
-The `language`-method should return which language to use on initialisation. This can either be a string or a Promise that resolves a
+The `language` method returns the language to be used on initialization. This can either be a string or a promise that resolves a
```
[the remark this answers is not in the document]
LBakkerTDS: 1. Restored the use of 'should'.
2. Globally replaced 'promise' with 'Promise'.

docs/plugins/language.md
```diff
@@ -1,45 +1,44 @@
...
   }
 
 }
 ```
 
-### Translations file
+### Translations File
 
-By default the Language plugin expects a `json` file, called `translations.json` in the `static` folder of your App.
+By default, the Language plugin expects a file with translations in JSON format, called **translations.json**, which is located in the **static** folder of your App.
```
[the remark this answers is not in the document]
LBakkerTDS: Changed text as follows: "An important prerequisite for using the Language plugin is a file with translations in JSON format, which is by default called **translations.json** and located in the **static** folder of your App.

> It is important that you create this file before using the Language plugin."

docs/plugins/language.md
```diff
@@ -147,42 +143,42 @@ Language.set(languageCode).then(() => {
...
+Setting the language is *instant* for translations specified inline in the **translations.json** file.
+For external language files, an *asynchronous* load request will be made.
 
-The language code should match with one of the language-keys specified in your `translations.json`.
+The language code should match with one of the language keys specified in your **translations.json**.
 
-The `map` option in the `meta` key of your translations file can be used to specify _aliases_, and as such can be used
-to map different ISO codes (i.e. en-GB, en-US, etc.) to a single language key (en).
+The `map` option in the `meta` key of your translations file can be used to specify *aliases*, and as such can be used
+to map different ISO codes (i.e., en-GB, en-US) to a single language key (en).
```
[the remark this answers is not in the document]
LBakkerTDS: Yep. See [https://www.instructionalsolutions.com/blog/how-to-use-i-e-and-e-g](https://www.instructionalsolutions.com/blog/how-to-use-i-e-and-e-g). 

By the way, the correct acronym here should be 'e.g.' instead of 'i.e.', because the languages are merely examples. Changed accordingly.

docs/plugins/lightning.md
```diff
@@ -1,18 +1,16 @@
 # Lightning
 
-The Lightning SDK is built on top of Lightning, but it also exposes Lightning as a module that can be imported in your App. In other words, there is no need to install Lightning-core as a separate dependency.
+The Lightning SDK is built on top of Lightning (Core), but it also exposes Lightning as a module that can be imported in your App. In other words, there is no need to install Lightning Core as a separate dependency.
```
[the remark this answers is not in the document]
LBakkerTDS: OK, first phrase deleted.

docs/plugins/lightning.md
```diff
@@ -1,18 +1,16 @@
...
-See <a href="https://rdkcentral.github.io/Lightning/" target="_blank">Lightning documentation</a> for more information.
+Apart from that, you can use Lightning to take advantage of the default texture and tools made available by Lightning (see [Using Tools from Lightning](#using-tools-from-lightning) below).
 
 ## Usage
 
-### Extending a Lightning component
+### Extending the Lightning Component
 
-All components in your app (including the main App) are required to extend a Lightning component.
+In addition to *importing* Lightning from the Lightning SDK, you must *extend* the Lightning component in *all* components of your App (including the main `App` component in the **App.js** file):
```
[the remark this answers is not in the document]
LBakkerTDS: Now that I see it, I agree. I have rewritten the sentence as follows: "It is required that *all* components in your App (including the main App) extend the Lightning component."

docs/plugins/pin.md
```diff
@@ -1,75 +1,86 @@
...
+Hides the visible **Pin** dialog.
 
 ```js
 Pin.hide()
 ```
 
-### Submit
+### submit
+
+Sends a Pin code to the middleware layer for verification. If the code is correct, the STB is unlocked.
```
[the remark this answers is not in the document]
LBakkerTDS: Agree. There is a difference in time between 'is' and 'will be'. The STB will only be unlocked after the PIN correctness check, so 'will be' is a better choice here.

docs/plugins/pin.md
```diff
@@ -1,75 +1,86 @@
...
-Pin.submit('0000').then(() => console.log('Unlocked!')).catch(e => console.log('Pin error', e))
+Pin.submit('0000')
+  .then(() => console.log('Unlocked!'))
+  .catch(e => console.log('Pin error', e))
 ```
 
-The `submit`-method returns a _promise_ which _resolves_ `true` when the supplied Pincode is correct and the STB is successfully unlocked.
-When the Pincode is wrong, the _promise_ will resolve `false`. In case the middleware is unable to unlock the STB, the promise will be
-_rejected_ (with an optional error message).
+If the supplied Pin code is correct, the `submit` method returns a *promise* which resolves with 'true' and the STB is successfully unlocked.
```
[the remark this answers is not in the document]
LBakkerTDS: You are absolutely right. I am sorry about this misinterpretation.;-( Text corrected as follows: "The `submit` method returns a *Promise*. If the supplied Pin code is correct, the Promise resolves with `true` and the STB will be unlocked. If the Pin code is wrong, the Promise resolves with `false`.

If the middleware is unable to unlock the STB, the Promise is *rejected* (with an optional error message)."

Global change "to resolve with <something>", not "to resolve <something>" or "to resolve to <something>".

docs/plugins/pin.md
```diff
@@ -1,75 +1,86 @@
...
+
+If the Pin code is wrong, the returned promise resolves with 'false'. If the middleware is unable to unlock the STB, the promise is
+*rejected* (with an optional error message).
 
-During development the default Pincode will be `0000`. Optionally you can overwrite the default Pincode during development by editing the
-`settings.json` file and adding the key `pin` as a `platformSetting` with a different valid Pincode.
+The default Pin code is  `0000`. Optionally, you can overwrite the default Pin code during development by editing the
```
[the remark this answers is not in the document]
LBakkerTDS: Agree, another misinterpretation. Text changed as follows: "During development, the default Pin code is  `0000`. Optionally, you can overwrite the default Pin code during development by editing the [...]".

docs/plugins/pin.md
```diff
@@ -1,75 +1,86 @@
...
-  .then(unlocked => unlocked === true ? console.log('STB is unlocked') : console.log('STB is locked'))
+  .then(
+    unlocked => unlocked === true ?
+      console.log('STB is unlocked') :
+      console.log('STB is locked'))
 ```
 
-The `unlocked`-method returns a _promise_ which _resolves_ `true` when the device is _unlocked_ and `false` when the device is _locked_.
-The promise is _rejected_ when the middleware is unable to retrieve the current state.
+The `unlocked` method returns a *promise* which resolves to 'true'  if the device is *unlocked* or to 'false' if the device is *locked*.
```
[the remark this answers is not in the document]
LBakkerTDS: Absolutely, I now understand why you use them. I will make this a global change.

docs/plugins/pin.md
```diff
@@ -1,75 +1,86 @@
...
+  .then(
+    locked => locked === true ?
+      console.log('STB is locked') :
+      console.log('STB is unlocked'))
 ```
 
-The `locked`-method is the exact counter part of the `unlocked`-method. It returns a _promise_ which _resolves_ `true` when
-the device is _locked_ and `false` when the device is _unlocked_.
-The promise is _rejected_ when the middleware is unable to retrieve the current state.
+The `locked` method is the *exact* counterpart of the `unlocked` method. It returns a *promise* which resolves to 'true' if
```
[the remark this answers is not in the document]
LBakkerTDS: See above.

docs/plugins/profile.md
```diff
@@ -1,150 +1,151 @@
 # Profile
 
-Sometimes your App might require profile information about the current user. This information generally comes from the platform or operator.
+Occasionally, your App requires profile information about the current user. This information is usually provided by the operator or platform.
 
-Since each operator / platform may have a different implementation, the Lightning SDK offers a Profile plugin with a generic interface to the developer.
+Because each operator or platform might implement user profile information in a different way, the Lightning SDK provides the *Profile* plugin. This plugin is a generic interface for developers, independent of any operator or platform.
 
-The Profile plugin also offers the ability to update profile information.
+You can also use the Profile plugin to *update* profile information.
```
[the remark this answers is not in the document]
LBakkerTDS: Note added.

docs/plugins/profile.md
LBakkerTDS: Text changed as follows: "If you pass parameters, the method will *attempt to update* the Profile information. The information that can actually be updated, depends on the operator / platform."

docs/plugins/profile.md
LBakkerTDS: Suggestion for "Gets the user's mother language." => "Gets the user's language, based on the currently set locale."

docs/plugins/language.md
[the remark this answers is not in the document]
michielvandergeest: cool .. every day is a school day :)

docs/plugins/purchase.md
LBakkerTDS: Typo corrected.

docs/plugins/purchase.md
LBakkerTDS: Sentence about custom `headers` added.

docs/plugins/purchase.md
LBakkerTDS: Example added.

docs/plugins/purchase.md
LBakkerTDS: Info added.

LBakkerTDS: I also noticed that the sentence "If the CSP backend [...]" was placed in the Setup section, where it should be in the Buy section. Sentence moved to the Buy section.

docs/plugins/purchase.md
LBakkerTDS: I have rewritten the sentence as: "The `household` is automatically sent as a `query param` in the request, which allows to optionally return information on whether the assets were purchased by the current household or not."

docs/plugins/purchase.md
LBakkerTDS: See above.

docs/plugins/purchase.md
LBakkerTDS: **Custom**: 'used before another word to mean "specially designed for a particular person or purpose":
custom-designed, custom-built software.' See 'adjective' section in [https://dictionary.cambridge.org/dictionary/english/custom](url).

**Customized**: 'made or changed according to the buyer's or user's needs'. See [https://dictionary.cambridge.org/dictionary/english/customized](url)

I think 'customized' is better replaced with 'custom-built' here.

docs/plugins/purchase.md
LBakkerTDS: Agreed.

docs/plugins/tv.md
```diff
@@ -1,35 +1,35 @@
 # TV
 
-The TV plugin serves as an abstraction layer to the _live TV_ functionality of a STB. The interface gives you access to the current channel
-and program information, and allows you to change the TV channel from inside an App (if your app is _whitelisted_ to do so).
+The *TV* plugin serves as an abstraction layer to the *live TV* functionality of an STB. The interface gives access to the current channel
+and program information, and allows you to change the TV channel from *inside* an App (if your App is *whitelisted* for that purpose).
 
-With the TV plugin you can adapt your App _contextually_ to what the user is currently watching on TV.
+You can use the TV plugin to adapt your App contextually to what the user is currently watching on TV.
```
michielvandergeest, line 6: This wording makes it sounds as if it's the _only_ thing you can do with the plugin. But it's _a_ thing you can do. The previous sentence maybe also doesn't cover that completely, by the way.

docs/plugins/tv.md
```diff
@@ -1,35 +1,35 @@
...
 
-#### Retrieve current Channel
+#### Retrieve Current Channel
 
 ```js
 TV.channel().then(channel => console.log(channel))
 ```
 
-The `channel`-method returns a _promise_ which returns the channel information as an object.
+The `channel` method returns a *Promise* that returns the channel information as an object.
```
michielvandergeest, line 28: Maybe we should add that the promise is rejected in case of an error (or if the app is now allowed to retrieve the information)

docs/plugins/tv.md
```diff
@@ -38,34 +38,34 @@ During _development_ the `channel`-method returns a random mocked channel. Optio
...
 
 ```js
 const channelNumber = 2
 TV.channel(channelNumber).then(channel => console.log(channel))
 ```
 
-When a channelNumber is passed as an argument to the `channel`-method it will try to change the TV channel.
-It will return a _promise_ which returns the channel information as an object.
+If a `channelNumber` is passed as an argument, the `channel` method attempts to change the TV channel.
+It returns a *Promise* that returns the channel information as an object.
```
michielvandergeest, line 49: Maybe we should add that the promise is rejected in case of an error (or if the app is now allowed to set the channel)

docs/plugins/versionlabel.md
```diff
@@ -1,7 +1,7 @@
...
+# VersionLabel
 
-During debug and QA it's important to know the _exact_ version of the App that's being tested.
+During debugging and QA, it is important to know the *exact* version of the App that is being tested.
 
-To facilitate this the SDK contains a built-in **Version Label** that can be turned on or off via the [Platform Setting](/plugins/settings?id=platform) `showVersion` in `settings.json`.
+The SDK contains a built-in *VersionLabel* which can be enabled or disabled via the [Platform Setting ](settings.md#platform-settings)`showVersion` in **settings.json**.
 
-When enabled an unubtrusive label with the App version as specified in `metadata.json` and the Lightning-SDK version will be displayed as an overlay in the bottom right corner of the App.
+If `showVersion` is enabled, an subtle label containing the App version (as specified in **metadata.json**) and the Lightning SDK version is displayed as an overlay in the bottom right corner of the App.
```
michielvandergeest, line 7: an subtle => a subtle

docs/plugins/videoplayer.md
```diff
@@ -1,31 +1,32 @@
...
+various [media events](#events) (for example: canplay, play, pause, seeking, seeked).
+
+> Although it is possible to implement a fully custom video playback solution, the use of the VideoPlayer plugin from
 the SDK is highly recommended.
 
 ## Usage
 
-In the Lightning components that require video playback capabilities (i.e., a Player component), you can import the
-VideoPlayer plugin from the Lightning SDK.
+In the Lightning components that require video playback capabilities (i.e., Player components), you can import the
```
michielvandergeest, line 16: I think I would keep this singular. An App usually has 1 Player component.

docs/plugins/videoplayer.md
```diff
@@ -37,30 +38,31 @@ class Player extends Lightning.Component {
...
 
 ### position
 
 Sets the x and y position of the video player.
 
-The `position` method accepts 2 arguments (`top` and `left`). Both values should be absolute numbers (either positive or negative).
-The default values for `top` and `left` are `0`.
+The `position` method accepts 2 arguments: `top` and `left`. Both values must be positive or negative absolute numbers. They both default to 0 (zero).
+
+For example:
```
michielvandergeest, line 51: I don't think we need 'for example' for each code example :)

docs/plugins/videoplayer.md
```diff
@@ -280,7 +285,7 @@ VideoPlayer.looped // false
 
 ### src
 
-Getter that retrieves the video player's current source (`src`).
+Getter that retrieves the video player's current source (src).
```
michielvandergeest, line 288: let's keep the backticks around src

docs/plugins/videoplayer.md
```diff
@@ -377,40 +375,51 @@ VideoPlayer.visible // false
 ## Events
 
 The VideoPlayer plugin emits a number of [media events](https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Media_events) to its
-consumer (as specified in [`VideoPlayer.consumer()`](#consumer)).
+*consumer* (which is specified in the [`VideoPlayer.consumer()`](#consumer) method).
```
michielvandergeest, line 378: maybe it should be "specified via the consumer method" ?

michielvandergeest (changes requested):
Added more feedback. The router part of the documentation is still pending.

docs/plugins/purchase.md
```diff
@@ -36,22 +38,20 @@ Purchase.buy(assetId).then(() => {
...
+The `setup` method accepts a *configuration object* as its only argument. The object can have *2 keys*:
 
-`endpoints` can be used to customize those API endpoints that differ from the default specification (as outlined below). For each _endpoint_ that deviates from the default, specify the _uri_ (which will be appended to the `cspUrl`, unless specified as a fully qualified domain) and the REST _method_ to use.
+* `cspUrl`: specifies the *base URL* of the backend of the CSP (Content Service Provider).
+* `endPoints`: can be used to customize those API endpoints that differ from the default specification (as outlined below). For each *endpoint* that differs from the default, you specify the *URI* (which will be appended to the `cspUrl`, unless specified as a fully qualified domain) and the REST *method* to use.
 
-If needed you can also specify an object with custom `headers` that will be merged in with the defaults headers (`{ Accept: 'application/json', 'Content-Type': 'application/json' }`)
+If needed, you an also specify an object with custom `headers` that will be merged with the default headers (`{ Accept: 'application/json', 'Content-Type': 'application/json' }`).
```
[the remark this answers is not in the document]
LBakkerTDS: Corrected.

docs/plugins/purchase.md
```diff
@@ -99,61 +99,33 @@ Purchase.setup('http://csp-backend.com/api', {
...
 ```
 
-In some cases the CSP backend setup is too different from the default specification, that it becomes too difficult
-to configure the enpoint uri's. In those cases you can specify a `callback` instead, and handle the request to the
-CSP backend entirely yourself.
+In some cases, the setup of the CSP backend differs too much from the default specification and it becomes too difficult to configure the enpoint URIs. In those cases, you can specify a *callback* instead, and handle the request to the CSP backend entirely yourself.
 
-The callback function will receive 2 arguments: `params` and `data` and should _always_ return a `Promise` which resolves the result (or rejects an error in case of a failure). When a callback is specified, all other keys in that `endpoint` config
-are ignored.
+The callback function will receive 2 arguments: *params* and *data*. It should always return a Promise which resolves with the result (or rejects with an error in case of a failure).
```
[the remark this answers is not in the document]
LBakkerTDS: Done. Apply also to _title_, _description_ and _price_ in "Typically, it returns an Array of objects containing information about the assets (such as 'title', 'description' and 'price')."?

docs/plugins/purchase.md
```diff
@@ -99,61 +99,33 @@ Purchase.setup('http://csp-backend.com/api', {
...
 })
 ```
 
-### Asset
+### asset
 
 The `asset` method returns a Promise with the result of the request to the CSP to retrieve details for a specific asset.
 
-The `household` is automatically sent as a `query param` in the request, allowing to return information on whether the asset has been _purchased_ by the current household.
+The `household` is automatically sent as a `query param` in the request. This allows to return information on whether the assets were purchased by the current household or not.
```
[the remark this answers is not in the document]
LBakkerTDS: You are absolutely right, corrected.

docs/plugins/purchase.md
```diff
@@ -99,61 +99,33 @@ Purchase.setup('http://csp-backend.com/api', {
...
 
-The `asset` method is an _optional_ convenience method exposed by the Purchase plugin. Depending on the CSPs API integration, the functionality to retrieve assets can be implemented fully custom in an App.
+The `asset` method is an *optional* convenience method provided by the Purchase plugin. Depending on the CSP's API integration, the functionality to retrieve assets can be fully custom-implemented in your App.
```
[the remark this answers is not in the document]
LBakkerTDS: This should be made consistent as well, corrected.

docs/plugins/purchase.md
```diff
@@ -99,61 +99,33 @@ Purchase.setup('http://csp-backend.com/api', {
...
-        headers: {
-          'Accept': 'application/json',
-				  'Content-Type': 'application/json'
-        }),
-        method: 'POST',
-      })
-    }
-  },
-})
-```
```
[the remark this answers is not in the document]
LBakkerTDS: Example restored between "> When a callback is specified, [...]" and "### assets".

docs/plugins/registry.md
```diff
@@ -1,37 +1,36 @@
...
-The Registry plugin can be used to globally register and unregister _event listeners_, _timeouts_ and _intervals_.
+You use the *Registry* plugin to globally register and unregister *event listeners*, *timeouts* and *intervals*.
 
 The plugin primarily serves as a proxy for the standard HTML5 Web APIs for setting and clearing listeners, intervals and timeouts
-(i.e. `window.setTimeout()` and `element.addEventListener()`).
+(for example, `window.setTimeout()` and `element.addEventListener()`).
 
-The added benefit of using the _Registry_ plugin is that any running timers, intervals and event listeners will be
-automatically cleaned up when closing the App.
+An additional benefit of the *Registry* plugin is that any running timers, intervals and event listeners are
```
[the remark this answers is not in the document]
LBakkerTDS: You are right, don't know why I changed that. Restored "The added benefit".

docs/plugins/registry.md
```diff
@@ -122,17 +118,21 @@ Registry.addEventListener(target, event, cb)
...
 
-The `removeEventListeners` method accepts 2 optional arguments (`target` and `event`):
+* If *no* arguments are passed, *all* previously registered listeners are removed.
 
-- When no arguments are passed, _all_ previously registered listeners will be removed
-- When passed only a `target` argument, all event listeners for the specified _target_ will be unregistered
-- When passed a `target` and an `event` argument, all listeners for that specific _event_ on that _target_ will be removed
+* If *only* a `target` argument is passed, all event listeners for the specified *target* are unregistered.
+
+* If a `target`*and* an `event` argument are passed, all listeners for that specific *event* on that *target* are removed.
```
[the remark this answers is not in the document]
LBakkerTDS: It was a list, only with more whitespace between the list items. Nevertheless I replaced '*' with '-' and removed the newlines between the items.

docs/plugins/settings.md
```diff
@@ -1,8 +1,8 @@
 # Settings
 
-An App created with the SDK can be configured with various options. These can either be _App_ or _Platform_ specific settings.
+An App that is created with the Lightning SDK, can be configured with various options. These options can be grouped into [App settings](#app-settings) and [Platform settings](#platform-settings).
```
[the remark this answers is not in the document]
LBakkerTDS: Agreed.

docs/plugins/settings.md
```diff
@@ -21,97 +21,100 @@ During development you can configure your own settings in `settings.json` to sim
...
-Optionally you can specify a `fallback`-value for when the setting is not defined.
+* `Type` can be either `app`, `platform` or `user`.
+* `Key` can be any of the existing settings.
+* You can specify an optional `fallback` value if the setting is not defined.
 
-### Set
+### set
 
-Sets a key, value on `user` type for _user defined settings_
+Sets a key for *a user-defined setting* where  `value` is of type `user`.
```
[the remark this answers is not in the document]
LBakkerTDS: Added note: "> The `set` method cannot overwrite App or Platform Settings."

docs/plugins/settings.md
```diff
@@ -21,97 +21,100 @@ During development you can configure your own settings in `settings.json` to sim
...
 
 ```js
 Settings.has(type, key)
 ```
 
-### Subscribe
+### subscribe
 
-Adds a callback to be notified when the value of a `user` key is changed. Multiple callbacks can be added for the same key.
+Adds a callback to the notification stack. The callback is notified when the value of a `user` key has changed. You can add multiple callbacks for the same key.
```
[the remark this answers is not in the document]
LBakkerTDS: Corrected.

docs/plugins/storage.md
```diff
@@ -1,57 +1,59 @@
...
+### set
 
 Saves a key-value combination in storage.
 
 ```js
 Storage.set(key, value)
 ```
 
-The key is expected to be a `String`. The value can be a `String`, `Object`, `Boolean` or `Array`. The value is automatically converted to a JSON object when saving, so there's not need to call `JSON.stringify()` on objects yourself.
+The `key` can be of type String (which is usually the case), Object, Boolean or Array.
```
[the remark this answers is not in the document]
LBakkerTDS: You are absolutely right, I somehow put the key and value types together. Sorry about that. I've reinstated the two separate sentences.

docs/plugins/storage.md
```diff
@@ -1,57 +1,59 @@
...
+### get
 
 Retrieves previously stored data from storage.
 
 ```js
 Storage.get(key)
 ```
 
-In case you stored an `Object`, the data is automatically converted back to an object, so there is no need to call `JSON.parse()` yourself.
+If you stored an Object, the data is automatically converted back to an Object, soyou do not have to call  `JSON.parse()`.
```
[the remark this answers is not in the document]
LBakkerTDS: Corrected.

docs/plugins/storage.md
```diff
@@ -1,57 +1,59 @@
...
 
 ```js
 Storage.remove(key)
 ```
 
-### Clear
+### clear
 
-Removes all data from LocalStorage
+Removes *all* data from localStorage.
```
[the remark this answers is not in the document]
LBakkerTDS: I was wondering why it mentioned 'LocalStorage' here and not under '###remove'. Corrected into 'storage'.

docs/plugins/tv.md
LBakkerTDS: Agreed. Changed the sentence into "One of the things you can do with the TV plugin, is adapt your App contextually to what the user is currently watching on TV."

docs/plugins/tv.md
LBakkerTDS: Added: "The Promise is *rejected* if an error occurred or if the App is not allowed to set the channel."

docs/plugins/versionlabel.md
LBakkerTDS: Corrected.

docs/plugins/videoplayer.md
LBakkerTDS: Corrected.

docs/plugins/videoplayer.md
LBakkerTDS: Okay!

docs/plugins/videoplayer.md
LBakkerTDS: Corrected.

docs/plugins/videoplayer.md
LBakkerTDS: Agreed.

docs/plugins/settings.md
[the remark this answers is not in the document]
LBakkerTDS: I also found "`Settings.clearSubscribers()" without the required ```js and ``` around it. Corrected.

docs/plugins/profile.md
michielvandergeest: yes, that sounds like a good change

docs/plugins/metadata
```diff
@@ -0,0 +1,75 @@
+# Metadata
+
+An App created with the SDK has a Metadata in it.
+
+Be aware that when your App is deployed to the App Store the Metadata is configured in the Store dashboard.
+
+## Usage
+
+Sometimes you may want to access Metadata in your App code directly. In those cases you can import the _Metadata plugin_ from the Lightning SDK.
```
michielvandergeest: Perhaps it will make it more clear if we provide an example? Something like: "... in your App code directly, for example to display the App version in about screen."

docs/plugins/metadata
```diff
@@ -0,0 +1,75 @@
...
+
+The SDK automatically creates the `appMetadata`  from the Launch params.
```
michielvandergeest: I think we can skip this one.

docs/plugins/metadata
```diff
@@ -0,0 +1,75 @@
...
+
+```js
+import { Metadata } from '@lightningjs/sdk'
+```
+
+## Available methods
+
+### Get
```
michielvandergeest: I would put this method last

michielvandergeest (approved):

PR #222, opened 2021-04-22
Added OS & SDK version fields

jlacivita:


mhughesacn:
Hi jlacivita: Please use Copyright ... Metrological instead of RDK in these two files.  We had to change the copyright holder line  so it expects this instead.
Thanks, Martin
Martin Hughes RDK CMF Compliance Team

jlacivita:
> Hi jlacivita: Please use Copyright ... Metrological instead of RDK in these two files. We had to change the copyright holder line so it expects this instead.
> Thanks, Martin
> Martin Hughes RDK CMF Compliance Team

Thanks!

Issue #189, opened 2021-02-16
Changes required in VideoTexture.js


magomez:
There are a couple of issues that are problematic in the code of VideoTexture.js. Both of them are around this piece of code:

https://github.com/rdkcentral/Lightning-SDK/blob/f89720ac765ff47497079852d0e9102b82af8c79/src/VideoPlayer/VideoTexture.js#L86-L90

The idea of that code is checking whether there is a new video frame ready to be rendered by using webkitDecodedFrameCount. If that value is not available, then it switches to render the video frame every 30ms.

The first problem is that webkitDecodedFrameCount is about to be deprecated. It should be replaced with the totalVideoFrames value defined in the new VideoPlaybackQuality interface (https://developer.mozilla.org/en-US/docs/Web/API/VideoPlaybackQuality). Ideally the code should support both systems at the same time to allow compatibility with older webkit versions.

The other problem is related to the fallback to paint every 30ms. This doesn't seem to be working. I haven't tested extensively, but I think the problem is that the variable this._lastTime is never given a value. Due to that, "this._lastTime < currentTime - 30" never evaluates to true. I think initializing the value to 0 should be enough to make it work.

erikhaandrikman:
Thanks for the report, we will pick this up

michielvandergeest:
Addressed in https://github.com/rdkcentral/Lightning-SDK/commit/c0f60d1b43d27bceb6e4839e0e316898881e2909

PR #224, opened 2021-04-26
Fix/video tecture issue

sandeep-vedam:
Fixed Video texture issue


michielvandergeest (approved):
LGTM

Should fix issue #189

PR #226, opened 2021-04-28
feat(FTL): Created Firebolt Transport Layer API

jlacivita:
## Overview
Created a Transport Layer API for sending/receiving JSON-RPC from a Firebolt Execution Environment
  
## Justification
The Transport Layer API should be agnostic to the actual Firebolt API, to minimize code duplication / chance of typos / etc.

## Changes
- Added a mock Transport Layer w/ a "defaults.js" in each module
- Created a two-way handshake to wire up SDK to FTL regardless of which loads first
- Created a Queue FTL, to store API requests until the handshake is done
- Created a way to store mock events in each defaults.js, so the mock FTL can dispatch them for testing

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,7 +2,7 @@
  * If not stated otherwise in this file or this component's LICENSE file the
  * following copyright and licenses apply:
  *
- * Copyright 2020 RDK Management
+ * Copyright 2021 Metrological
  *
  * Licensed under the Apache License, Version 2.0 (the License);
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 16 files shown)


mhughesacn:
Hi @jlacivita The headers on this repository had to be switched to Metrological, so the three index.js files that have RDK Management copyrights on line 5 need to switch to Metrological instead.
Thanks,
Martin Hughes, RDK CMF Compliance Team

cletustboone (approved):

jlacivita:
Thanks again @mhughesacn !

PR #228, opened 2021-04-30
fix #227: Null check in dotGrab

### LINKED ISSUE #227: Settings not checking for Null values
https://github.com/rdkcentral/Lightning-SDK/blob/779e85aff338f812b411ab4cbcd7fe2fd7a1b9bd/src/Settings/index.js#L38

Doesn't check for null which is an object

chiefcll:


michielvandergeest:
Closing as this has been fixed in v4.3.2, same solution effectively, just one memory allocation less ;)

PR #241, opened 2021-06-10
Fix/readme open course

sfloam:
I believe the README.md should say open source instead of open course.


michielvandergeest (approved):
Thanks! :)

PR #244, opened 2021-06-17
Feature/router

erikhaandrikman:


CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Changelog
 
+## v4.x.x
```
michielvandergeest, line 3: Let's make it 4.4.0 (since there's a new feature)? And add a release date

michielvandergeest (changes requested):

CHANGELOG.md
erikhaandrikman: yeah, i've flagged it as a new feature but it was actually a bug fix. Updated changelog with version and release date

michielvandergeest (approved):

PR #253, opened 2021-07-23
Pin context

sandeep-vedam:
Following are the code changes in this PR :
1. Updated Pin related code to support for 'parental' or 'purchase' in submit, locked and unlocked methods
2. Updated documentation accordingly as per the changes in the code


src/Pin/index.js
```diff
@@ -23,21 +23,25 @@ import { ApplicationInstance } from '../Launch'
 
 // only used during local development
 let unlocked = false
+let contextItems = ['purchase', 'parental']
```
michielvandergeest, line 26: we're not reassigning `contextItems`, so this should be a constant :)

src/Pin/index.js
```diff
@@ -23,21 +23,25 @@ import { ApplicationInstance } from '../Launch'
...
 
-let submit = pin => {
+let submit = (pin, context) => {
   return new Promise((resolve, reject) => {
-    if (pin.toString() === Settings.get('platform', 'pin', '0000').toString()) {
+    if (
+      pin.toString() === Settings.get('platform', 'pin', '0000').toString() &&
+      contextItems.includes(context)
+    ) {
```
michielvandergeest, line 33: We shouldn't perform the valid `context` check here, since this function is overwritten by the platform plugin. Instead it should be added to the public API `submit`-method, before we pass it to the actual `submit` method.

This will allow to throw a specific log about the invalid context. Context should still be passed to the platform submit method though

src/Pin/index.js
```diff
@@ -23,21 +23,25 @@ import { ApplicationInstance } from '../Launch'
...
     }
   })
 }
 
-let check = () => {
-  return new Promise(resolve => {
-    resolve(unlocked)
+let check = context => {
+  return new Promise((resolve, reject) => {
+    contextItems.includes(context) ? resolve(unlocked) : reject('Incorrect Context')
```
michielvandergeest, line 44: Same goes for this method. The valid context check should be in the public API methods, before reaching the platform check function

src/Pin/index.js
```diff
@@ -73,32 +77,32 @@ export default {
     )
     pinDialog = null
   },
-  submit(pin) {
+  submit(pin, context = contextItems[0]) {
```
michielvandergeest, line 80: It would be nice if we can advise the developer that it's recommended to pass an explicit `context`. This means that we should remove the default value here, and assign it inside the function if left empty (and add a `Log.info()`)

michielvandergeest (changes requested):

src/Pin/index.js
```diff
@@ -73,34 +83,46 @@ export default {
...
     pinDialog = null
   },
-  submit(pin) {
+  submit(pin, context) {
     return new Promise((resolve, reject) => {
       try {
-        submit(pin)
-          .then(resolve)
-          .catch(reject)
+        if (contextCheck(context) || contextItems.includes(context)) {
```
michielvandergeest, line 89: I would move the check for valid context (`contextItems.includes(context)`) to the `contextCheck` function. That makes it a bit cleaner to read, plus it's more DRY (since this includes check exists also in `unlocked`).

michielvandergeest (changes requested):

michielvandergeest (approved):
LGTM

PR #263, opened 2021-09-11
Router: Allow periods in hash values

frank-weindel:
Given the route defintion:
```
/route/:param
```
And given the hash:
```
/route/this.is.my.value
```

Currently Lightning SDK router does not return the entire expected value for `:param` and instead only returns the value as `'this'`.

This PR allows periods to be in a hash value.

QUESTION: Is there a reason this is so restrictive? Could we perhaps change the regex to something much less restrictive like?:

```
/(\/?:?[^/]+)/g
```


erikhaandrikman:
There is no real reasoning behing only allowing `@!*\w%\s:-` it just grew towards this set over time. It's even missing other accepted (unreserved characters) https://datatracker.ietf.org/doc/html/rfc3986#section-2.3

We will put the investigation of allowing all accept forward slash `[^/]` on our backlog

erikhaandrikman (approved):

Issue #292, opened 2022-01-20
Bug - using a 'beforeEachRoute' and 'beforeNavigate' on a route results in unexpected behaviour


robbertvancaem:
**Version of SDK:**
4.6.1

**Steps to reproduce**
- Define a global `beforeEachRoute` function in the router configuration which returns `false`
- Define a `beforeNavigate` function on a specific route which logs to the console

**Expected behaviour**
- There is no log to the console

**Actual behaviour**
- The `beforeNavigate` function logs to the console

I believe this is due to the result of the `beforeEachRoute` call being overridden by the result of the `beforeNavigate` call. According to the [documentation](https://lightningjs.io/docs/#/lightning-sdk-reference/plugins/router/configuration?id=beforeeachroute), returning `false` should stop the router from navigating immediately, and therefore should not call the `beforeNavigate` method. 

See these lines:
https://github.com/rdkcentral/Lightning-SDK/blob/6d239403a40c4f53ea1054ebd79124e712102ea1/src/Router/index.js#L252-L256

michielvandergeest:
This fix has been released today (v4.8.2)

PR #297, opened 2022-02-10
Beforeach route fix

sandeep-vedam:
Updated code to make sure beoreEachRoute has priority

Fix for #292


michielvandergeest (approved):
Looks good to me.

@erikhaandrikman can you double check we're indeed fixing a bug here (and introducing a new one or breaking intended behaviour with this change)

Issue #313, opened 2022-04-11
Router doesn't automatically close App


mauro1855:
Hello,

According to the Router documentation (particularly [here](https://lightningjs.io/docs/#/lightning-sdk-reference/plugins/router/events)), an App that implements the router is supposed to automatically exit when the back key is pressed and the route history is empty.

> When the Router’s history is empty, the SDK will continue to handle the Back key and close the App. You can prevent this by adding _handleAppClose() to your App class, for instance to show an Exit dialog.

However, this is currently not happening. Even when the route history is empty, the app doesn't automatically close. I've discussed with @thomvl and @erikhaandrikman and we believe this to be a bug. 
For example, if you look at the [router-example-app](https://github.com/mlapps/router-example-app), you'll see that in order to make the app close the creator was forced to put a ```_handleAppClose()``` implementation and explicitly tell the app to close.

Can you confirm that this is a bug and that the Router should do this automatically?

**Steps to reproduce:**
1. Create a App that implements the Router (or use the [router-example-app](https://github.com/mlapps/router-example-app))
2. Make sure to NOT implement the ```_handleAppClose()```
3. Start the app and press the back key right after.

**Expected behaviour:**
The app should close since the router history is empty.

**Current behaviour:**
Nothing happens.

michielvandergeest:
Hi @mauro1855, this indeed looks like something that should be fixed.

Thanks for your PR, we'll look into your fix! 👍

michielvandergeest:
Merged into `dev`-branch. The fix will become available in the next release.

PR #314, opened 2022-04-11
CloseApp automatically when no more router history

mauro1855:
A potential fix for #313, assuming it is indeed a bug.


michielvandergeest (approved):
LGTM

PR #317, opened 2022-04-26
Fix named route property not being accessed properly

MLangendijk:
This is a fix for the Named Route navigation paradigm as documented here: https://lightningjs.io/docs/#/lightning-sdk-reference/plugins/router/navigation?id=named-routes

The fix adds a `getter method` on the `Route` model for the `name` property. The `getRouteByName` function in the route utils checks for the name property to retrieve the correct route. Since this property wasn't available on the model, it would never find the correct route and never execute any. Reference: https://github.com/rdkcentral/Lightning-SDK/blob/dev/src/Router/utils/route.js#L240

### DIFF
--- a/src/Router/model/Route.js
+++ b/src/Router/model/Route.js
@@ -39,6 +39,10 @@ export default class Route {
     return this._cfg.path
   }
 
+  get name() {
+    return this._cfg.name
+  }
+
   get component() {
     return this._cfg.component
   }


MLangendijk:
@michielvandergeest I somehow can't add reviewers; I'm assuming this needs to be reviewed though?

michielvandergeest:
Hi @MLangendijk, thanks for your contribution! Yeah, I'll make sure this get's properly reviewed and tested this week so we can merge and release it if all is good.

sandeep-vedam:
Hey @MLangendijk ,
We’ve validated the fix you have provided and it looks good. We will merge it and release it with the next release

MLangendijk:
> Hey @MLangendijk , We’ve validated the fix you have provided and it looks good. We will merge it and release it with the next release

Thanks for the update!

michielvandergeest (approved):
LGTM @MLangendijk thanks for your contribution.

PR #318, opened 2022-05-12
feat: add support of async Launch call in index.js

mikita-liaukovich-epam:
This PR adds support of asynchronous app launching.

Example of `index.js` (*notice `async\await` usage*):
image


mikita-liaukovich-epam:
Closing as for this comment https://github.com/rdkcentral/Lightning-SDK/pull/132#issuecomment-712490263:
`startApp.js mimicks the Metrological App Platform during local development. So while this change works locally, it won't work in the context of the App store.`

PR #319, opened 2022-05-16
Added base 64 image support for Image plugin

sandeep-vedam:
Added base 64 image support for Image plugin


michielvandergeest (approved):
LGTM

Issue #326, opened 2022-06-24
VideoTexture position animation is inconsistent


mauro1855:
Hi Lightning SDK team,

I created recently a little project to see how I could create "video thumbnails" that start playing when I focus the element - it involved punching a hole in the canvas (with the HoleShader) and changing the position and size of the VideoPlayer (with `VideoPlayer.position()` and `VideoPlayer.size()` methods)

But I found that the `VideoPlayer.position()` behaviour is not consistent when `textureMode` is turned on or off. 
* If textureMode is turned off, the style is updated and the video player tag "jumps" to the location.
* If textureMode is turned on, the texture is "smoothed" into the location, thus causing a delay until it actually reaches there. This means that, with the a hole punched in the canvas, you can see the video texture sliding into place.

I find this behaviour difference between texture and non-texture mode to be inconsistent, and provides an unpredictable and uncontrollable experience to developers. Not only with the `VideoPlayer.position()` method, but with the `.size()`, `.show()` and `.hide()` methods as well - the VideoTexture animates / smooths all of them, while the video tag is directly assigned the CSS style which takes effect immediately without animation. Besides being inconsistent, it is rather opinionated as well; it assumes the developer wants the smooth effect, while it might not be the case.

You can find the relevant code [here](https://github.com/rdkcentral/Lightning-SDK/blob/805784b1361148b36deea9f8b12a00eb834e22e4/src/VideoPlayer/index.js#L203) and [here](https://github.com/rdkcentral/Lightning-SDK/blob/93c749509bb9a5db5e2888b8eed7d76ab3ac72ec/src/VideoPlayer/VideoTexture.js).

Would you consider a PR to remove the smoothness out of the video texture?

Best regards,
Mauro

michielvandergeest:
Hi @mauro1855, thanks for your report.

Starting v5.0.0 the VideoPlayer plugin is no longer part of the Lightning-SDK. It’s been moved to the [Metrological-SDK](https://github.com/Metrological/metrological-sdk).

I’m not able to move issues between organizations, so I would suggest to open this issue as a new ticket [over here](https://github.com/Metrological/metrological-sdk/issues).

PR #327, opened 2022-06-24
Remove opinionated smooth behavior from VideoTexture

mauro1855:
Hi team,

As mentioned in #326, the VideoPlayer plugin behaviour is currently inconsistent when running in textureMode vs. non-textureMode. Particularly, the inconsistency lies in the immediate effect of CSS styling vs the delayed and animated effect of the smoothed property updates in textureMode.
Additionally, a developer using the SDK has no control over the animation parameters, which makes the current implementation very opinionated.

I hope you don't find this change too radical, since the objective is to make the behaviour of the VideoPlayer consistent in both textureMode and non-textureMode and provide a more predictable developer experience.

Best regards,
Mauro

### DIFF
--- a/src/VideoPlayer/VideoTexture.js
+++ b/src/VideoPlayer/VideoTexture.js
@@ -147,27 +147,23 @@ export default class VideoTexture extends Lightning.Component {
 
   position(top, left) {
     this.videoView.patch({
-      smooth: {
-        x: left,
-        y: top,
-      },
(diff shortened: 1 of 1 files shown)


michielvandergeest:
Thanks for your PR @mauro1855.

The PR itself looks good to me though! So I would suggest to recreate this PR at the metrological-sdk to get it merged in there.

Issue #328, opened 2022-06-28
playAfterSeek in VideoPlayer not working


mauro1855:
Hi SDK team,

The VideoPlayer plugin currently has some logic on `seek()` to resume video playback if it was already playing when `seek()` was called. You can find the logic [here](https://github.com/rdkcentral/Lightning-SDK/blob/master/src/VideoPlayer/index.js#L322) and then [here](https://github.com/rdkcentral/Lightning-SDK/blob/master/src/VideoPlayer/index.js#L76).

Unfortunately this is currently not working - the playback doesn't resume. I checked with @bjornvanspengen and @tomvdbroek and we believe the issue lies in the fact that the `play()` is called before the video player is ready to play after the seek. 

I think the player needs to wait for the `canplay` player event before it can play. I will be submitting a PR with this solution soon.

michielvandergeest:
Hi @mauro1855, thanks for your report.

Starting v5.0.0 the VideoPlayer plugin is no longer part of the Lightning-SDK. It’s been moved to the [Metrological-SDK](https://github.com/Metrological/metrological-sdk).

I’m not able to move issues between organizations, so I would suggest to open this issue as a new ticket [over here](https://github.com/Metrological/metrological-sdk/issues).

PR #329, opened 2022-06-28
Wait for canplay event before playing after seek

mauro1855:
PR to address #328

This solution doesn't work in Safari, but the previous code also didn't work there, so nothing lost.
I checked a couple of STBs and confirmed this addresses the issue of not resuming playback after seek.

### DIFF
--- a/src/VideoPlayer/index.js
+++ b/src/VideoPlayer/index.js
@@ -73,7 +73,8 @@ const hooks = {
     state.playing = false
   },
   seeked() {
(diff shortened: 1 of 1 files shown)


michielvandergeest:
Thanks for your PR @mauro1855.

The PR itself looks good to me, only point of attention is to make sure the event listener gets properly removed from the video element upon app close.

I would suggest to recreate this PR at the metrological-sdk to get it merged in there.

PR #333, opened 2022-07-05
VideoPlayer: capitalize Error event to maintain consistency

ricardo-a-alves-alb:
Every event that is fired on the component that is consuming a `VideoPlayer` event is capitalized, resulting in a method named something like `$videoPlayerPlay`.

However, for the Error event, it is not capitalized properly which means that the method would be named something like `$videoPlayererror`.

This PR attempts to address this inconsistency.

### DIFF
--- a/src/VideoPlayer/events.js
+++ b/src/VideoPlayer/events.js
@@ -18,9 +18,12 @@
  */
 
 export default {
+  adstart: 'AdStart',
+  adend: 'AdEnd'
   abort: 'Abort',
   canplay: 'CanPlay',
   canplaythrough: 'CanPlayThrough',
+  clear: 'Clear',
   durationchange: 'DurationChange',
   emptied: 'Emptied',
   encrypted: 'Encrypted',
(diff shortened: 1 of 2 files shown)


erikhaandrikman:
Since `error` is a property of the `events` object: https://github.com/rdkcentral/Lightning-SDK/blob/82b8948ca0e7efe7afe5a94ba5700d21a9cc880c/src/VideoPlayer/events.js#L28

is suggest using that value so we can maintain all event names in one place: 
`fireOnConsumer(events['error'], { videoElement: videoEl, event: e })`

ricardo-a-alves-alb:
> Since `error` is a property of the `events` object:
> 
> https://github.com/rdkcentral/Lightning-SDK/blob/82b8948ca0e7efe7afe5a94ba5700d21a9cc880c/src/VideoPlayer/events.js#L28
> 
> is suggest using that value so we can maintain all event names in one place: `fireOnConsumer(events['error'], { videoElement: videoEl, event: e })`

Just changed it and did the same to the `Clear`, `AdStart` and `AdEnd` events.

michielvandergeest:
Thanks for this PR @ricardo-a-alves-alb 

I would suggest to open this PR there to see if it can get merged in.

PR #337, opened 2022-07-29
TypeScript Support

frank-weindel:
This adds partial TypeScript support and TSDoc to Lightning SDK. The focus areas of these types are as follows
- Router plugin
- Image plugin
- Utils plugin
- Platform settings
- Typing for the Lightning export is available in PR rdkcentral/Lightning#407
- Any untyped areas have been stubbed with `any`

_**No runtime code files (.js / .mjs) are modified or added in this PR**_

### Automated Testing

- See rdkcentral/Lightning#407

### API Docs + Generation

- See rdkcentral/Lightning#407

### Still to-do in this PR

- [x] Wait for release of rdkcentral/Lightning#407 and set the minimum version of the Lightning Core dependecy to that version.

### Still to do in future PRs

- See rdkcentral/Lightning#407

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -94,5 +94,9 @@ export interface PlatformSettings {
      * for more information.
      */
     quality?: number
+    showFps?: boolean;
+    showVersion?: boolean;
+    log?: boolean;
+    path?: string;
   }
 }
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,3 +1,5 @@
+import { RouterPlatformSettings } from './src/Router/index.js';
+
 /*
  * If not stated otherwise in this file or this component's LICENSE file the
  * following copyright and licenses apply:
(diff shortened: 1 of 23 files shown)


index.d.ts
```diff
@@ -0,0 +1,159 @@
...
+export const Metadata: any; // export { default as Metadata } from './src/Metadata'
+
+/**
+ * [Platform Settings](https://lightningjs.io/docs/#/lightning-sdk-reference/plugins/settings?id=platform-settings)
+ */
+export interface PlatformSettings {
+  plugins?: {
+    profile?: any;
+    metrics?: any;
+    mediaPlayer?: any;
```
michielvandergeest, line 54: the mediaPlayer plugin is deprecated and we really don't recommend it being used anymore (it's also removed from the docs already). Should we maybe not include it here?

frank-weindel: It's typed as `any` here so we're not really giving the user any guidance on how its used. I can remove it though

frank-weindel:
Updated with minimum dependency of version 2.7.0 of Core

michielvandergeest (approved):
👍

Issue #348, opened 2022-09-22
Router UpdateHash not working


mauro1855:
Hi Lightning SDK team,

As far as I can tell, the `updateHash` router setting is not currently doing anything. No matter if I set it to true or false, the hash is always updated in the URL.

**Steps to replicate**:
1. Download the [router-example-app](https://github.com/mlapps/router-example-app)
2. Change router settings in the `settings.json` file so that `"updateHash": false`
3. Start the app with `lng dev`.

**Expected behaviour:**
The URL hash should not be updated when routes change. (is that correct?)

**Observer behaviour:**
The URL hash is updated when the route changes. Changing the `updateHash` property has no effect.

michielvandergeest:
Hi @mauro1855 thanks for this report.

We have indeed confirmed that the updateHash option doesn't have any effect.

@sandeep-vedam can you look into a fix for this?

michielvandergeest:
This one has been merged and will be released in our upcoming October release.

PR #349, opened 2022-10-03
Fix/update hash

sandeep-vedam:
Fix for Update Hash https://github.com/rdkcentral/Lightning-SDK/issues/348

With this fix : 
updateHash set to True  -  hash will be updated
updateHash set to False - hash wont be updated

### DIFF
--- a/src/Router/index.js
+++ b/src/Router/index.js
@@ -182,9 +182,7 @@ export const navigate = (url, args = {}, store) => {
 
     if (mustUpdateLocationHash()) {
       setHash(url)
-    }
-
-    if (!mustUpdateLocationHash()) {
+    } else {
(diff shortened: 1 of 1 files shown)


src/Router/index.js
```diff
@@ -180,7 +180,10 @@ export const navigate = (url, args = {}, store) => {
     // push request in the queue
     queue(url, args, store)
 
-    setHash(url)
+    if (mustUpdateLocationHash()) {
+      setHash(url)
+    }
+
     if (!mustUpdateLocationHash()) {
```
michielvandergeest, line 187: let's make this into an `else`-statement

michielvandergeest (approved):
LGTM

PR #361, opened 2022-11-24
ENH: Allow plugins to load local json files

### LINKED ISSUE #360: ENH: Allow plugins to load local json files
On certain plugins, such as the `Language` plugin, when deployed to devices that use the `file:///` protocol, if we ship our App with a local translation file, it fails on some older platforms when using the `fetch` API polyfill.

For example, on Tizen 3.0 that has the Chromium version 39, it uses the `fetch` API polyfill and fails to load the translations file causing our App to stay on a black screen.
However, on Tizen 5.5 that has the Chromium version 69 and doesn't need to polyfill the `fetch` API, it loads with no issue.

I'll provide a PR for you to take a look at and merge if it seems good to you.

ricardo-a-alves-alb:
On certain plugins, such as the `Language` plugin, when deployed to devices that use the `file:///` protocol, if we ship our App with a local translation file, it fails on some older platforms when using the `fetch` API polyfill.

For example, on Tizen 3.0 that has the Chromium version 39, it uses the `fetch` API polyfill and fails to load the translations file causing our App to stay on a black screen.
However, on Tizen 5.5 that has the Chromium version 69 and doesn't need to polyfill the `fetch` API, it loads with no issue.

closes #360

### DIFF
--- a/src/Colors/index.js
+++ b/src/Colors/index.js
@@ -17,6 +17,7 @@
  * limitations under the License.
  */
 
+import fetchJson from '../helpers/fetchJson'
 import Log from '../Log'
 import { mergeColors, calculateAlpha, isObject, isString, argbToHSLA, hslaToARGB } from './utils.js'
 
@@ -60,8 +61,7 @@ export const initColors = file => {
       addColors(file)
       return resolve()
     }
-    fetch(file)
-      .then(response => response.json())
+    fetchJson(file)
       .then(json => {
         addColors(json)
         return resolve()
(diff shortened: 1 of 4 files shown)


michielvandergeest:
Thanks for this contribution @ricardo-a-alves-alb! We haven't run into this issue before with Chrome39, but possibly Tizen has a slightly different implementation.

The change you made looks good and we've tested the functionality, so we're good to merge this. We'll release this with our next quarterly release (planned for January).

PR #362, opened 2022-12-02
Added router support for optional path

sandeep-vedam:
Added support for optional path in Router

### DIFF
--- a/src/Router/utils/router.js
+++ b/src/Router/utils/router.js
@@ -263,14 +263,20 @@ const init = config => {
     }
   }
   config.routes.forEach(item => {
-    const path = stripRegex(item.path)
+    // replacing regexes with 'R' to avoid issues with pattern matching below
+    const strippedPath = stripRegex(item.path)
+
     // Pattern to identify the last path of the route
     // It should start with "/:" + any word  and ends with "?"
-    // Example /player/:asset/:assetId?
+    // It should be the last path of the route
+    // valid => /player/:asset/:assetId? (:assetId is optional)
+    // invalid => /player/:asset/:assetId?/test (:assetId? is not an optional path)
+    // invalid => /player/:asset?/:assetId? (second path is not considered as an optional path)
     const pattern = /.*\/:.*?\?$/u
-    if (pattern.test(item.path)) {
-      const optionalPath = path.substring(0, path.lastIndexOf('/'))
-      const originalPath = path.substring(0, path.lastIndexOf('?'))
+
+    if (pattern.test(strippedPath)) {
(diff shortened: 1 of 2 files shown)


ricardo-a-alves-alb:
How would the route look like?

`player/:id?`

michielvandergeest (commented):
LGTM - can we just change some `let`s into `const`s as discussed.

frank-weindel:
Seems a bit fragile to me and would really benefit from documentation.

One might assume that a path like this should work:

```
'player/:assetId?/:playlistId'
```

But it would result in these two Route paths being used without any indication of error:

```
'player/:assetId'
'player/:assetId?'
```

frank-weindel:
FWIW I found the library that Express uses for Route path parsing. Wondering if it might help: https://www.npmjs.com/package/path-to-regexp

anandpatel0123:
Hi, when this is going to be rolled out? eagerly waiting.

uguraslan:
The router plugin supports using regular expressions in route paths (as documented [here](https://lightningjs.io/docs/#/lightning-sdk-reference/plugins/router/configuration?id=using-regular-expressions-in-routes)) like:

```js
// this will match #player/1493847
// but will fail on #player/ah26134
{
    path: 'player/:playlistId${/[0-9]{3,8}/i}',
    component: Player
}
```

`?` is part of regex syntax, it can be even used as a literal character in regex (`\?`). The following example will fail due to the current optional path implementation.

```js
{
    path: 'player/:playlistId${/[0-9]{3,8}0?/i}',
    component: Player
}
```

src/Router/utils/router.js
```diff
@@ -262,6 +262,22 @@ const init = config => {
       console.error(`[Router]: ${config.bootComponent} is not a valid boot component`)
     }
   }
+  config.routes.forEach(item => {
+    let path = item.path
+    // Pattern to identify the last path of the route
+    // It should start with "/:" + any word  and ends with "?"
+    // Example /player/:asset/:assetId?
+    const pattern = /\/:{1}\w+\?$/
```
uguraslan, line 270: `\w` matches `[A-Za-z0-9_]` character class, whereas as far as I see there is no restriction about path names in the Router docs. For example, they can include hyphen `-` or and other characters from other languages. Also `{1}` is not required when we try to match single literal character. So I think the following regex might be better:

```js
const pattern = /\/:.*?\?$/u
```

uguraslan:
@michielvandergeest I think everything seems to be okay for this PR. Can you please review it again with the latest changes?

michielvandergeest (approved):
LGTM

PR #364, opened 2022-12-02
Over ride keymap

sandeep-vedam:
Added support for overriding the keymap with the new keymap


src/Application/index.js
```diff
@@ -119,6 +119,22 @@ export default function(App, appData, platformSettings) {
...
 
+    get keymapping() {
+      return this.stage.application.config.keys
+    }
+
+    set keymapping(overrideKeymap) {
+      const baseKeymap = this.stage.application.config.keys
+      Object.keys(overrideKeymap).reduce((keymapping, key) => {
+        // prevent duplicate values to exist in final keymapping (i.e. 2 keys listening to 'Back')
+        const duplicateIndex = Object.values(baseKeymap).indexOf(overrideKeymap[key])
```
uguraslan, line 130: It is currently possible to define 2 different keys listening to 'Back' by using `settings.json`. Using 2 different keys for the same handler is also referenced in [src/Application/index.js](https://github.com/rdkcentral/Lightning-SDK/blob/2b0d10e6d36ae9d855d8f0ac29e8eaccae807ebf/src/Application/index.js#L55). 

I think it could be better to reconsider removing duplicated values here while overriding keys because duplicated values might be coming from `settings.json`.

uguraslan: As an alternative way, `overrideKeymap` can completely replace the existing `baseKeymap` and the decision of merging key map can be left to developers. Because with the overriding feature they can provide any key map (either merged with the base key map or a completely new key map) based on their needs.

michielvandergeest: As discussed offline: we’ll keep the default behaviour as is. We will however introduce a flag to enable keeping defaults (or disable the removal of defaults, whichever way way of looking at it is more logical).

This means that the setter will have to become a function (which makes more sense anyway) in order to accept the option flag.

src/Application/index.js
```diff
@@ -123,16 +123,28 @@ export default function(App, appData, platformSettings) {
...
-
-        keymapping[key] = overrideKeymap[key]
-        return keymapping
-      }, baseKeymap)
+    /**
+     * This function overides the default keymap with the latest keymap.
+     * @param overrideKeymap
+     * @param keepDefaults
+     */
+    overRideKeyMap(overrideKeymap, keepDefaults = false) {
```
michielvandergeest, line 131: I would rename the flag to refer to `duplicates` instead of `defaults` (i.e `keepDuplicates` / `removeDuplicates` / `allowDuplicates`)

uguraslan (approved):
LGTM

michielvandergeest (approved):

PR #366, opened 2022-12-14
fix: Router set previous state when using the on data provider

### LINKED ISSUE #365: Router: using on provider changes the app state to the page instead of maintaining on the previous
Hello,

We were implementing a menu as a widget and came across a weird bug when using the `on` data provider on a route.
Each menu item does a `Router.navigate('path')` and, when we use the `on` data provider the focus would be stolen by the page instead of maintaining the focus on the page.

After some digging, I found the culript.

Here, it is setting and exporting a `previousState` variable:

https://github.com/rdkcentral/Lightning-SDK/blob/82b8948ca0e7efe7afe5a94ba5700d21a9cc880c/src/Router/utils/provider.js#L25-L32

Here, it is using a local variable with public get/set:

https://github.com/rdkcentral/Lightning-SDK/blob/eeb784375fa28e66286411997609d3302af4e97e/src/Router/utils/router.js#L147-L152
https://github.com/rdkcentral/Lightning-SDK/blob/eeb784375fa28e66286411997609d3302af4e97e/src/Router/utils/router.js#L439-L445

However, the setter method is never called because the `provider.js` file is using a different variable instead of the setter.

https://github.com/rdkcentral/Lightning-SDK/blob/2b0d10e6d36ae9d855d8f0ac29e8eaccae807ebf/src/Router/utils/loader.js#L53-L61

I'll provide a PR to fix this issue.

ricardo-a-alves-alb:
fixes #365

### DIFF
--- a/src/Router/utils/provider.js
+++ b/src/Router/utils/provider.js
@@ -18,15 +18,13 @@
  */
 
 import { symbols, getQueryStringParams } from './helpers'
-import { app, routes, routeExists } from './router'
+import { app, routes, routeExists, setPreviousState } from './router'
 import { getValuesFromHash } from './route'
 import emit from './emit'
 
-export let previousState
-
 export const dataHooks = {
   on: request => {
-    previousState = app.state || ''
+    setPreviousState(app.state || '')
     app._setState('Loading')
     return execProvider(request)
   },


ricardo-a-alves-alb:
@michielvandergeest maybe you could take a look at this? Thanks!

michielvandergeest (approved):
Thanks for this contribution! LGTM

PR #371, opened 2023-01-30
Feature/subtitle plugin

jfboeve:
Fix visibility property issue
Added viewport width and height getters and setters and incorporated into calculations
Added functions to index file

### DIFF
--- a/src/Subtitles/SubtitleComponent.js
+++ b/src/Subtitles/SubtitleComponent.js
@@ -81,7 +81,7 @@ export default class SubtitleComponent extends Lightning.Component {
     this.x = this._calculateX(this.xPos)
   }
 
-  get viewPortW() {
+  get viewportW() {
     return this._viewportW || this.application.finalW
   }
 
@@ -96,11 +96,11 @@ export default class SubtitleComponent extends Lightning.Component {
 
   _calculateX(x) {
     if (x === 'center') {
-      x = (this.viewPortW - this.finalW) / 2
+      x = (this.viewportW - this.finalW) / 2
     } else if (x === 'left') {
       x = 60
     } else if (x === 'right') {
-      x = this.viewPortW - this.finalW - 60
+      x = this.viewportW - this.finalW - 60
     }
     return x
   }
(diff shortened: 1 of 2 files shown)


src/Subtitles/SubtitleComponent.js
```diff
@@ -137,11 +153,12 @@ export default class SubtitleComponent extends Lightning.Component {
 
   _defineBreakpoint(text, breakpoint) {
     if (breakpoint >= this.maxWidth) return this.maxWidth
+    const textureDefaults = new Lightning.textures.TextTexture(this.stage).cloneArgs()
```
michielvandergeest, line 156: I agree it seems more logical to put this functionality in the method that uses it, but I'd say it's more efficient to keep this line in the `_init()` method so it's only executed once (and not every time the breakpoint is calculated - which happens multiple times)

src/Subtitles/SubtitleComponent.js
```diff
@@ -162,9 +179,7 @@ export default class SubtitleComponent extends Lightning.Component {
 
       this.tag('Text').text.wordWrapWidth = breakpoint
       this.tag('Text').text = v
-      setTimeout(() => {
-        this.setSmooth('alpha', 1, { duration: 0.2 })
-      }, 100)
+      this.alpha = 1
```
michielvandergeest, line 182: 👍 as discussed .. not need to have an animation here.

src/Subtitles/index.js
```diff
@@ -75,6 +76,27 @@ export default {
...
+    subtitles.viewportH = h
+  },
+  xPos(v) {
+    const subtitles = getOrCreateSubtitlesComponent()
+    subtitles.xPos = v
+  },
+  yPos(v) {
+    const subtitles = getOrCreateSubtitlesComponent()
+    subtitles.yPos = v
+  },
```
michielvandergeest, line 99: wouldn't it be more intuitive to have only one method for this that accepts an object with 4 keys (width, height, x and y)?

michielvandergeest (changes requested):
see inline comments

src/Subtitles/SubtitleComponent.js
jfboeve: Oh right, forgot to revert that back!

src/Subtitles/index.js
jfboeve: From a frontend point of view I really like the individual value method, this allows you to change the values a bit more dynamically during run time without you having to know f.e. the width of the viewport. While using methods with more parameters are handy during the initialisation phase of the player.
The individual value methods are also very handy when using the `styles` method, this allows you to do: 

Subtitles.styles({
  xPos: 'center',
  fontSize: 40,
  viewportH: 540
})

PR #368, opened 2023-01-11
Fix Application version label positioning

frank-weindel:
Fixes
- #359

PR #353, opened 2022-10-06
Feature/subtitle plugin

sandeep-vedam:
Changes related to Subtitle plugin.

PR #373, opened 2023-02-03
Release - v5.3.0

### LINKED ISSUE #276: Override key mapping after app is launched
Hello,
is it possible to setup the key mapping **after** the app is launched? Reading the lightning sdk, it seems the keys register can happen only inside the Application class' constructor, no way to invoke this with a public method.

### LINKED ISSUE #350: Allow optional parameters to be specified in the Lightning Router path
As of now with the latest SDK, there is no option in router path to specify optional parameters.

For example: if I want to make the playlistId, an optional field in the below path, it would not be possible.
{
  path: 'player/:assetId/:playlistId',
  component: Player
}

I would either need to specify two different routes to the same component.

It would be very helpful to have a way to specify optional fields in router path with a '?' suffix for the optional params.

example:
{
  path: 'player/:assetId/:playlistId?',
  component: Player
}

would route to

http://some/url#player/1
http://some.url#player/1/2

### LINKED ISSUE #359: App version wrongly positioned
Hi Lightning SDK team,

Since Lightning SDK 5.2.0, the app version is wrongly position for apps that are set to run in 1280x720 (0.6667 precision). See here a screenshot:

image

**Steps to reproduce:**
1. Create a new Lightning app with the CLI (`lng create`)
2. Add the following `stage` settings to the `setting.json` file:
```
"w": 1280,
"h": 720,
"precision": 0.66667
```

**Expected behaviour:**
The app version's blue rectangle should be positioned in the bottom right corner, as it has been in previous Lightning SDK versions.

**Observed behaviour:**
The app version's blue rectangle positioning is not correct.

**Happens since:** 5.2.0

### LINKED ISSUE #360: ENH: Allow plugins to load local json files

### LINKED ISSUE #365: Router: using on provider changes the app state to the page instead of maintaining on the previous

### LINKED ISSUE #375: Backtracking docs with no matching code
Under deeplinking - there is a backtracking flag
https://rdkcentral.github.io/Lightning-SDK/#/plugins/router/deeplinking?id=backtracking

However, there is no actual code based on that flag. So either something needs adding to the code or the docs need removing.

https://forum.lightningjs.io/t/backtracking-is-not-working-deeplinking/573

uguraslan:
## Changes

**Merged**

- Fixed #360 with [PR #361 and PR #370]
- Fixed #365 with [PR #366]
- Fixed #359 with [PR #368]
- Updated `@metrological/sdk` and `localCookie` NPM package paths with [PR #369]
- Fixed #350 with [PR #362]
- Added Subtitles plugin for displaying subtitles (closed captions) with [PR #353]
- Fixed #276 with [PR #364]
- Fixed #375 with [PR #376]

## To-do

- [x] Update `CHANGELOG.md`
- [x] Update `docs/changelog.md`


uguraslan:
All PRs have been merged into the release branch of v5.3.0. This PR is ready for the final review before the release.

src/Application/index.js
```diff
@@ -119,6 +119,32 @@ export default function(App, appData, platformSettings) {
...
+    get keymapping() {
+      return this.stage.application.config.keys
+    }
+
+    /**
+     * This function overides the default keymap with the latest keymap.
+     * @param customKeymap
+     * @param keepDuplicates
+     */
+    overRideKeyMap(customKeymap, keepDuplicates = false) {
```
ricardo-a-alves-alb, line 131: Why is the word override capitalized like this?

src/Subtitles/SubtitleComponent.js
```diff
@@ -0,0 +1,203 @@
...
+    return {
+      visible: false,
+      rect: true,
+      color: 0x90000000,
+      shader: { type: Lightning.shaders.RoundedRectangle, radius: 5 },
+      Text: {
+        y: 5,
+        x: 20,
+        text: {
+          fontColor: 0xffffffff,
```
ricardo-a-alves-alb, line 33: This doesn't seem correct. Isn't the property `textColor`? [Related docs](https://github.com/rdkcentral/Lightning/blob/master/docs/RenderEngine/Textures/Text.md)

ricardo-a-alves-alb (changes requested):
Also, why was `static/fonts/Roboto-Regular.ttf` added?

sandeep-vedam:
> Also, why was `static/fonts/Roboto-Regular.ttf` added?

Thanks for noticing and highlighting this. @ricardo-a-alves-alb 

This was added at commit 56b8b93. The font was used in Subtitle plugin at the initial stages of development. Now there is no use for it as the feature is matured enough to use. We will remove this.

src/Application/index.js
uguraslan: Thanks @ricardo-a-alves-alb for pointing out that, we have updated the function name as well as some variables names to make it all consistent.

src/Subtitles/SubtitleComponent.js
uguraslan: Thanks, this also has been fixed.

src/Subtitles/SubtitleComponent.js
```diff
@@ -0,0 +1,203 @@
...
+    const textTag = this.tag('Text').text
+    return {
+      fontFace: textTag.fontFace || 'sans-serif',
+      fontSize: textTag.fontSize,
+      lineHeight: textTag.lineHeight,
+      textAlign: textTag.textAlign,
+      wordWrap: true,
+      maxLines: textTag.maxLines,
+    }
+  }
```
ricardo-a-alves-alb, line 66: Should the textColor be also returned here?

src/Subtitles/SubtitleComponent.js
```diff
@@ -0,0 +1,203 @@
...
+
+  set fontSize(v) {
+    this.tag('Text').text.fontSize = v
+    this.tag('Text').text.lineHeight = v * 1.3
+  }
+
+  set fontColor(v) {
+    this.tag('Text').color = v
+  }
```
ricardo-a-alves-alb, line 150: Should this also be renamed to textColor? And it should also patch the textColor, not the color property of the Texture.

src/Subtitles/SubtitleComponent.js
```diff
@@ -0,0 +1,203 @@
...
+
+  _init() {
+    this._textTextureDefaults = new Lightning.textures.TextTexture(this.stage).cloneArgs()
+
+    this.tag('Text').on('txLoaded', ({ _source }) => {
+      this.w = _source.w + this.tag('Text').x * 2
+      this.h = _source.h
+      this.position()
+    })
```
ricardo-a-alves-alb, line 53: Probably should add a TODO to fix this when this PR gets pushed to the core:
https://github.com/rdkcentral/Lightning/pull/451

src/Subtitles/index.js
```diff
@@ -0,0 +1,105 @@
...
+    subtitles.fontFamily = v
+  },
+  fontSize(v) {
+    const subtitles = getOrCreateSubtitlesComponent()
+    subtitles.fontSize = v
+  },
+  fontColor(v) {
+    const subtitles = getOrCreateSubtitlesComponent()
+    subtitles.fontColor = v
+  },
```
ricardo-a-alves-alb, line 66: Remains fontColor or is changed to textColor?

docs/plugins/subtitles.md
```diff
@@ -0,0 +1,144 @@
...
+Subtitles.fontSize(50)
+```
+
+### fontColor
+
+Sets the `fontColor` style of the Subtitles text
+
+```js
+Subtitles.fontColor(0xffffffff)
+```
```
ricardo-a-alves-alb, line 88: If you opt to change this property to textColor, just a reminder to also change the DOCS.

docs/plugins/subtitles.md
```diff
@@ -0,0 +1,144 @@
...
+
+Clears the text of the Subtitles text.
+
+```js
+Subtitles.clear()
+```
+
+### styles
+
+Sets the styles of the Subtitles text. This is a short hand method for setting the following styles: `fontFamily`, `fontSize`, `fontColor`, `backgroundColor`, `textAlign`. If any of these styles are not set, the default values will be used.
```
ricardo-a-alves-alb, line 55: Also here.

docs/plugins/subtitles.md
```diff
@@ -0,0 +1,144 @@
...
+
+```js
+Subtitles.styles({
+  fontFamily: 'sans-serif',
+  fontSize: 45,
+  fontColor: 0xffffffff,
```
ricardo-a-alves-alb, line 61: And also here.

ricardo-a-alves-alb (changes requested):

src/Subtitles/SubtitleComponent.js
michielvandergeest: There is no need for the text color to be returned here. It's not relevant in the pre-render calculation of the text texture. We're only interested in the size of the texture and the color doesn't affect that

src/Subtitles/SubtitleComponent.js
michielvandergeest: It's actually less CPU intensive to change the color property of the texture versus setting the textColor of the text itself. I don't think there is a real need to change the name of the setter either.

src/Subtitles/SubtitleComponent.js
michielvandergeest: Thanks for pointing this out! We indeed have this on our internal backlog to verify with a future release, depending on when / whether PR 451 is merged.

docs/plugins/subtitles.md
uguraslan: The same explanation regarding `fontColor` is also valid here : https://github.com/rdkcentral/Lightning-SDK/pull/373#discussion_r1107165496

docs/plugins/subtitles.md
uguraslan: The same explanation regarding `fontColor` is also valid here : https://github.com/rdkcentral/Lightning-SDK/pull/373#discussion_r1107165496

docs/plugins/subtitles.md
uguraslan: The same explanation regarding `fontColor` is also valid here : https://github.com/rdkcentral/Lightning-SDK/pull/373#discussion_r1107165496

src/Subtitles/SubtitleComponent.js
ricardo-a-alves-alb: That's actually a great tip that never crossed my mind! 👍

ricardo-a-alves-alb: However, with this in mind, should the initial state also use the color and not the textColor property of the texture?

https://github.com/rdkcentral/Lightning-SDK/blob/246e700e4c19c7f0ca1f22fb255df2223fb0d721/src/Subtitles/SubtitleComponent.js#L29-L42

michielvandergeest: We could technically omit the initial textColor, since it default to white anyway, but I don't mind having it there for clarity reasons. The technique used (setting the color instead of the textColor) means that we are tinting the text Texture with the color. And since the text itself is white it will result in the exact color that we pass it (tinting a gray text texture would have a different result for example)

michielvandergeest (approved):
LGTM

PR #374, opened 2023-02-09
Application: Add missing "default default" for precision

frank-weindel:
This fixes a bug introduced by #368 where it was incorrectly assumed that precision was guaranteed to be set with a default in all cases. But apparently it was only specific cases.


michielvandergeest (approved):
LGTM

uguraslan (approved):
LGTM

PR #384, opened 2023-03-22
add announcer plugin

chiefcll:


mhughesacn:
Hi Chris: This is failing because it needs credit.  Where did you get the code in utils.js?

(Credit is somewhat complicated.  I think the following is a good source and would go under the Metrological Apache header.

Code from: https://github.com/jashkenas/underscore is
Copyright (c) 2009-2022 Jeremy Ashkenas, Julian Gonggrijp, and DocumentCloud and Investigative Reporters & Editors
Licensed under the MIT License
based off:
http://unscriptable.com/2009/03/20/debouncing-javascript-methods/ which is:
Copyright (c) 2007-2009 unscriptable.com and John M. Hann
Licensed under the MIT License

NOTICE should have the same text appended.

LICENSE needs a copy of the MIT license (with Copyright YEAR COPYRIGHT HOLDER  i.e. the generic fields)


Thank you,
Martin

chiefcll:
Hopefully followed your instructions correctly... LMK

LICENSE
```diff
@@ -201,3 +201,10 @@
    limitations under the License.
 
 
+Copyright 2023 Metrological
```
mhughesacn, line 204: Thanks Chris:
Sorry, I didn't make this clear: Please use the generic form not specific year etc. so the same entry will satisfy all MIT needs. i.e.

```
MIT License
===========
Copyright <YEAR> <COPYRIGHT HOLDER>

Permission is...
```

mhughesacn (changes requested):

NOTICE
```diff
@@ -7,3 +7,10 @@ listed below. Your use of this material within the component is also subject to
 conditions of these licenses. The LICENSE file contains the text of all the licenses which apply
 within this component.
  
+Code from: https://github.com/jashkenas/underscore is
+Copyright (c) 2009-2022 Jeremy Ashkenas, Julian Gonggrijp, and DocumentCloud and Investigative Reporters & Editors
+Licensed under the MIT License
+based off:
+http://unscriptable.com/2009/03/20/debouncing-javascript-methods/ which is:
+Copyright (c) 2007-2009 unscriptable.com and John M. Hann
+Licensed under the MIT License
```
mhughesacn, line 16: Hi Chris: I am really sorry about this but I just noticed that Mr.Hann says that his code is licensed under the MIT license but he doesn't use a conventional MIT license.  There is an extra clause when you scroll to the end.  I think it is enough to change line 16 to say:
`Licensed under the MIT License (with X11 advertising exception)
`Please change this line in utils.js as well.

mhughesacn (approved):

jfboeve:
Hey Chris,

I checked out the changed code and its looking a lot better! There are a few tiny things that need some adjusting:

- Move everything Announcer related to a separate folder within the Accessibility folder. (Don't forget to adjust the index.js to the src path)
- Move _textToSpeech out of the Announcer `const`, it seems like it's meant to be private.

michielvandergeest:
@sandeep-vedam as discussed: can you please do some testing with this PR to validate if everything works as expected and evaluate if the documentation is clear enough. Thanks :)

jfboeve (approved):

LICENSE
```diff
@@ -201,3 +201,10 @@
    limitations under the License.
 
 
+Copyright <YEAR> <COPYRIGHT HOLDER>
```
michielvandergeest, line 204: What does this copyright notice refer to exactly?

src/Accessibility/Announcer/utils.js
```diff
@@ -0,0 +1,95 @@
...
+ * that is a function which will clear the timer to prevent previously scheduled executions.
+ *
+ * @source underscore.js
+ * @see http://unscriptable.com/2009/03/20/debouncing-javascript-methods/
+ * @param {Function} function to wrap
+ * @param {Number} timeout in ms (`100`)
+ * @param {Boolean} whether to execute at the beginning (`false`)
+ * @api public
+ */
+export function debounce(func, wait, immediate) {
```
michielvandergeest, line 45: isn't this one available as an installable npm package, instead of copying over the code? Alternatively we could use the lodash equivalent, which I'm pretty sure has separate packages for each individual function

LICENSE
mhughesacn: Hi Michiel: This is an appendix to the generic Apache-2.0 license (actually where it tells you how to apply the license to code).  The NOTICE file is where explicit values are filled in.  We agreed this approach with Comcast Legal because the alternative meant the duplication of the text of the license for each copyright holder. - Martin

sandeep-vedam:
Hey @chiefcll 

I have tested this Announcer plugin with sample application(Router) and it seems to be fine. 

Thanks, 
Sandeep.

PR #386, opened 2023-03-31
Fixed router back track issue

sandeep-vedam:
Fixed issue related to Router backtrack. Here is the issue

When the router is configured with backtrack as true and the user enters App via a deeplink, clicking on the handle back results in either closing or showing an exit dialog if the app has.

This fix will serve when the user enters the deeplink and user presses the back button:
1. Backtrack: enabled ->  Application recursively removes the last part of the hash until it finds a valid path to navigate. In case there is no valid hash then it should exit the application.
2. Backtrack: disabled -> Application is closed

### DIFF
--- a/src/Router/index.js
+++ b/src/Router/index.js
@@ -357,14 +357,8 @@ export const step = (level = 0) => {
   // for now we only support negative numbers
   level = Math.abs(level)
 
-  // we can't step back past the amount
-  // of history entries
-  if (level > history.length) {
-    if (isFunction(app._handleAppClose)) {
-      return app._handleAppClose()
-    }
-    return app.application.closeApp()
-  } else if (history.length) {
+  //Check whether we have any history avaialble or not
+  if (history.length) {
     // for now we only support history back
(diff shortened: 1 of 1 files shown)


ricardo-a-alves-alb:
How about when we have the `root` property defined? Shouldn't it navigate to the root page before triggering the `_handleAppClose` callback?

imtiaz101325:
Hey @sandeep-vedam,
I'm also with @ricardo-a-alves-alb here. We have recently found this behavior and and have added a handler on the root component to solve this. Some apps might want to close the app when hitting back like this.

erikhaandrikman:
> How about when we have the `root` property defined? Shouldn't it navigate to the root page before triggering the `_handleAppClose` callback?

@ricardo-a-alves-alb this does sound like a good suggestion. For now we're going to merge this PR as is, since it's fixes the broken backtrack functionality.  We should probably discuss your suggested change for the next release

erikhaandrikman (approved):

PR #388, opened 2023-04-24
DeepMerge failed to merge instantiated objects

suresh-gangumalla:
I have observed that the DeepMerge library fails to correctly merge instantiated objects, which is causing issues with the Stage object context option. The canvas option value is being reassinged after DeepMerge but the context option is not being handled properly.

To resolve this issue, I have introduced a new option called 'isMergeableObject' in the DeepMerge library. This option copies the whole instantiated object, instead of just copying its properties. This allows the proper merging of instantiated objects and resolves both the canvas and context options issues with DeepMerge.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
         "@metrological/sdk": "^1.0.0",
         "@michieljs/execute-as-promise": "^1.0.0",
         "deepmerge": "^4.2.2",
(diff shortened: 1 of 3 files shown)


michielvandergeest (approved):
LGTM. Thanks for implementing this fix @suresh-gangumalla

PR #390, opened 2023-04-29
Kill previously keepAlive page instance

jay-patel9:
Router: Kill previously keepAlive page instance

Here is the scenario -

Page1 -> Page2(keepAlive:true) -> Page3 -> Page4 -> Page2(same instance will be used)

The fix includes replacing the used instance with the instance created at the startup of the App.

Page1 -> Page2(keepAlive:true) -> Page3 -> Page4 ->  Page2(new instance after calling deletePage(Page2))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## v5.3.2
+
+*28 apr 2023*
+
+- Fixed issue where some stage options are not properly preserved during deep merge (i.e canvas and webgl context) ([#388](https://github.com/rdkcentral/Lightning-SDK/pull/388))
+- Fixed router backtrack issue ([#385](https://github.com/rdkcentral/Lightning-SDK/issues/385)
+
 ## v5.3.1
 *23 feb 2023*
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## v5.3.2
+
+*28 apr 2023*
+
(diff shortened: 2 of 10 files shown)


sandeep-vedam:
Hello @jay-patel9 

Thanks for the PR. Regarding this fix I thoroughly tested the provided fix. It functions perfectly in cases where the route path doesn't involve any route parameters. However, I encountered a minor setback when the route contains parameters. Unfortunately, the fix doesn't seem to address this specific scenario.

As an example, let's consider the following routes:

{
  path: 'viewall/:category',
  component: ViewAll,
  widgets: ['AppBackground']
},
{
  path: 'video',
  component: VideoHome,
  widgets: ['AppBackground'],
}
During testing, I tried the following approach:

Router.deletePage('viewall') - Unfortunately, this method didn't work.
Router.deletePage('viewall/:category') - However, this approach did work successfully.

It appears that we need to provide the exact route path as specified in the routes.js file when using the deletePage function. I'm unsure whether this behavior is intentional or if it indicates a potential issue. It might be worth investigating further to determine the expected behavior in this situation.

sandeep-vedam:
Hey @erikhaandrikman , @michielvandergeest , @uguraslan ,

Please let us know your thoughts on this

erikhaandrikman:
Hi @jay-patel9 i suggest going for option 2 and use the full route to delete the page.

erikhaandrikman (approved):

michielvandergeest (approved):
Looks good me to as well 🙂

PR #392, opened 2023-07-04
Added extra eager cleanup of event listeners and targets to prevent r…

michielvandergeest:
Implemented extra eager cleanup in registry plugin to prevent any possible memory leak / memory retention during app lifetime.

Should fix an issue reported on Discord (https://discord.com/channels/1020300788016353311/1124249672890994689/1124249672890994689):

> If I interpret the code correctly, if any event is added, the target on which the event is listened will be kept indefinitely in the targets array of the registry?
> Does this not lead to memory usage buildup? Say you have video tracks on which you listen for events. Every video will have a different video track. Over time after longer usage of an app, the number of videotracks kept in the targets array could build up.

### DIFF
--- a/src/Registry/index.js
+++ b/src/Registry/index.js
@@ -130,6 +130,16 @@ export default {
         handler.toString()
       )
       target.removeEventListener(event, handler)
+      // remove key from event listeners object when no events are registered for that event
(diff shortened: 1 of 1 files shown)


sandeep-vedam (approved):
I have tested this changes and it works fine.
LGTM

PR #394, opened 2023-07-14
add Colors type definitions

ben-wall:


index.d.ts
```diff
@@ -22,7 +22,23 @@ export const Ads: any; // export { default as Ads } from './src/Ads'
...
+  mix(color: string, percentage: number): ColorTransform;
+  get(): number;
+};
+
+export type Colors = {
+  (colorName: string): ColorTransform;
+} & {
+  [colorName: string]: ColorTransform;
+};
+export const Colors: Colors; // export { default as Colors } from './src/Colors'
```
frank-weindel, line 41: Thanks for the PR! I'd happily be willing to help push it forward. Could you please move the type definitions to `src/Colors/index.d.ts` though? This file is meant to just meant to re-export types from each individual SDK module.

sandeep-vedam (approved):
LGTM

ben-wall:
I was hoping this would go through pretty fast so that I could continue to add types in the future and start using them without continuously patching/unpatching things as we develop our side, is there still something blocking this?

index.d.ts
frank-weindel: LGTM thanks!

PR #399, opened 2023-08-07
allow Colors to take ARGB values

WarrenMoore:
Currently, you can't specify the alpha property of a colour directly (instead you have to use the `.alpha()` method). This PR would allow you to use ARGB values directly.


sandeep-vedam (approved):
LGTM

jfboeve (approved):

PR #420, opened 2024-06-03
updated Colors type definition

uguraslan:
fixes the following error:

```
Error: Type 'typeof import("node_modules/@lightningjs/sdk/src/Colors/index")' has no call signatures
```


michielvandergeest (approved):
LGTM

